PR #1, opened 2020-05-14
Add ansible playbook to set up site

djgalloway:
We'll obviously need to replace `beta.ceph.io` with `ceph.io` in the future but this is working for now with `ansible-playbook -e branch=develop run.yml`

### DIFF
--- a/ansible/README.rst
+++ b/ansible/README.rst
@@ -1,0 +1,67 @@
+Deployment Playbook
+===================
+
+This ansible role can be used to configure/reconfigure the ceph.io website.
+
+Prerequisites
+-------------
+
+- Ubuntu 20.04 server with a public IP
+- nginx is the only supported webserver
+- Run the ``users``, ``common``, and ``public_facing`` roles from the ceph-cm-ansible_ repo.  This is optional but is an easy way to set up user accounts, SSH keys, nagios monitoring, firewall, and fail2ban.
+
+.. _ceph-cm-ansible: https://github.com/ceph/ceph-cm-ansible
+
+Variables
+---------
+
+I'd be surprised if this playbook got reused for any purpose other than setting up ceph.io but if desired, override these vars set in ``defaults/main.yml``
+
++----------------------------------------------+-----------------------------------------------------------------------------------------------+
+|                                              | Description                                                                                   |
+| Variable                                     |                                                                                               |
++==============================================+===============================================================================================+
+| ``fqdn: beta.ceph.io``                       | Fully qualified domain name of webserver.                                                     |
++----------------------------------------------+-----------------------------------------------------------------------------------------------+
+| ``install_dir: /opt/www``                    | Directory in which to clone and serve the ceph.io static site repo.                           |
++----------------------------------------------+-----------------------------------------------------------------------------------------------+
+| ``branch: master``                           | Branch of this repo to serve.  This is useful for testing site changes.                       |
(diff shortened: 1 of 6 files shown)


adamduncan:
Thanks for making ground on this @djgalloway! Happy to take your lead on deployment setup.

Had started drafting some questions for the team, which tie into this nicely: Which/how many hosting environments do we foresee wanting? Will these need securing, if so how? Who will be responsible for custom domains, configuring DNS records etc.?

This followed on from some assumptions we might consider for a project like this. Would be great to get your thoughts:

> The solution needs to be deployable via Git-linked continuous deployment. This should be both automated as part of the commit and merge request process, as well as achievable manually.
>
> Multiple environments will likely be required to offer development, staging/UAT and production sites, to support content and update workflows. Having the ability to preview commit or branch deploys would be beneficial for drafting, internal reviews and sign-off. Environments should be secured via SSL.
>
> Keeping ongoing hosting costs low (or nil if possible) is also important.

djgalloway:
> Had started drafting some questions for the team, which tie into this nicely: Which/how many hosting environments do we foresee wanting?

I'm satisfied with there only being one server.  beta.ceph.io is currently running on a highly available VM in a RHV cluster managed by me.

> Will these need securing, if so how?

That's taken care of by the ansible roles I mention in the README.

The [public_facing](https://github.com/ceph/ceph-cm-ansible/tree/master/roles/public_facing/) ansible role, specifically, makes sure `PasswordAuthentication no` in sshd_config.  It also configures UFW for firewalling.  As a bonus, `fail2ban` is deployed to block SSH bruteforce bots and can easily be adapted to monitor nginx logs for nefarious actors.

SSH access is managed by the [users](https://github.com/ceph/ceph-cm-ansible/tree/master/roles/users) role.

> Who will be responsible for custom domains, configuring DNS records etc.?

The ceph.io domain is owned by Red Hat.  @dmick @zmc and myself have push access to update that zone file.

> 
> This followed on from some assumptions we might consider for a project like this. Would be great to get your thoughts:
> 
> > The solution needs to be deployable via Git-linked continuous deployment. This should be both automated as part of the commit and merge request process, as well as achievable manually.

Yeah, the cronjob I have set up currently is clunky.  We already have a public Jenkins instance that could handle updating the site for us as well as testing dev changes.

> > Multiple environments will likely be required to offer development, staging/UAT and production sites, to support content and update workflows. Having the ability to preview commit or branch deploys would be beneficial for drafting, internal reviews and sign-off. Environments should be secured via SSL.

Should be fairly easy to set up a separate directory on the server and have nginx serve it up at `dev.ceph.io` or something similar.  Just trying to avoid spreading everything out for what's supposed to be a simple website :smile:

SSL is already set up and forced (http redirects to https) using the ansible role in this PR.

> > Keeping ongoing hosting costs low (or nil if possible) is also important.

As I mentioned above, the current beta.ceph.io is running in our upstream test lab so it's essentially free.

adamduncan:
Awesome, thank you for the detailed response 👍 That's super helpful, and looks like you've already got those points covered. Just wanted to make sure we were asking the right questions/had the right assumptions from our side.

djgalloway:
I'm going to merge this as-is; I consider it a decent starting point.  I'm going to work on getting a Jenkins agent process running on beta.ceph.io and having `https://$branch.ceph.io` point to build built by a Jenkins job.

PR #2, opened 2020-05-20
Support wildcard subdomains

djgalloway:
This will let us see a live/public view of $GIT_BRANCH.ceph.io when a branch is pushed.


djgalloway:
Siiiick!  @adamduncan we don't need to have it create a comment if it's too verbose.  But it works!

PR #10, opened 2020-11-25
Hotfix/collections fallbacks

adamduncan:
Just a couple quick fixes, spotted digging into this multilingual challenge:

1. Collections filters can have default fallbacks to ensure they behave as expected when not present
2. The en-GB blog names its tag collections pages `categories.html` — bringing ja-JP in line to match (currently `tags.html`)


Pete-Robelou (approved):
All looks good.

Good spot with the `tags`, `categories` naming. Obviously I missed it in the `ja-JP` site.

PR #12, opened 2021-01-19
sticky scroll and off canvas scripts, site header icons, site header …

Pete-Robelou:
…layout and CSS, nav primary layout and CSS, move secondary nav into main body.


src/_includes/components/navigation-secondary.njk
```diff
@@ -14,19 +17,22 @@
...
-      {% if page.url | startsWith(item.url) %}
-        <li><a class="text-semibold" href="{{ item.url | url }}">{{ item.title }}</a></li>
-        {%- if item.children %}
-          {%- for subitem in item.children %}
-            <li>
-              <a href="{{ subitem.url | url }}" {% if page.url | startsWith(subitem.url) %} aria-current="page"{% endif %}>{{ subitem.title }}</a>
-            </li>
-          {%- endfor %}
+{# Not sure about this if layout != home as it feels a bit flimsy #}
+{% if layout != 'home' %}
```
adamduncan, line 21: Yeah, it's a bit of an odd structure the navigation plugin gives us here, but hierarchically good. 

Could we perhaps check that `secondaryNav` isn't null and/or has a length, and then determine if its secondary nav (ie. `primaryNav[0].children`) has length/subitems to iterate over?

src/assets/svgs/icon-close.svg
```diff
@@ -0,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
```
adamduncan, line 1: Thought: Do we benefit from having `role="img"` or `aria-hidden="true"` on non-meaningful inline SVGs, to ensure they're not announced to assistive tech? [Ref](https://css-tricks.com/accessible-svgs/)

src/css/component.nav-primary.css
```diff
@@ -0,0 +1,70 @@
...
+    color: var(--color-grey-500);
+  }
+
+  .home .nav-primary a:hover,
+  .home .nav-primary a:active {
+    color: var(--color-white);
+  }
+
+  .home .site-header--stuck .nav-primary a {
+    color: var(--color-ggunmetal);
```
adamduncan, line 54: Spelling: `ggunmetal`

src/css/component.site-header.css
```diff
@@ -0,0 +1,107 @@
...
+}
+
+@media (--mq-md) {
+  .site-header {
+    --site-header-menu-gutters-x: var(--size-7);
+  }
+}
+
+@media (--mq-to-lg) {
+  .site-header__menu {
```
adamduncan, line 74: Thought: Could this do with `overflow-y: auto` (and are there scrolling implications for how the nav list items might scroll beneath the Ceph logo)? Thinking how this panel's contents will increase with language selector in future.

src/js/off-canvas.js
```diff
@@ -0,0 +1,48 @@
+"use strict";
+/*global document window*/
+
+(function() {
+
+  const html = document.querySelector('html');
+  const header = document.querySelector('.site-header');
+  const menu = document.querySelector('.site-header__menu');
+  const menu_open = document.querySelector('.site-header__menu-open');
+  const menu_close = document.querySelector('.site-header__menu-close');
```
adamduncan, line 10: Thought: Wonder if we can devise some semantic selector approach for grabbing DOM elements, which might be more tied to their role (or decoupled from their classnames, e.g. `.js-site-header__menu-open` or `[data-site-header-menu-open]`). What do you think?

src/js/off-canvas.js
```diff
@@ -0,0 +1,48 @@
...
+
+  const media_query = window.matchMedia('(width < 64em)')
+
+  function toggleMenu() {
+    menu_open.addEventListener('click', (e) => {
```
adamduncan, line 15: Suggestion (non-blocking): We could maybe pick these event handler functions out and refactor this a little to make it easier to also unbind the click events (should the user resize from small to large) 🤙

adamduncan (approved):
Looking good, @Pete-Robelou! I think let's push this out, and can always revisit and refine as we go.

src/css/component.site-header.css
```diff
@@ -0,0 +1,107 @@
...
+}
+
+.home .site-header--stuck {
+  background: var(--color-white);
+  border-bottom: var(--size-px) solid var(--color-grey-500);
+  box-shadow: 0 1rem 0.625rem -0.5rem hsla(var(--color-black-hsl), 0.05);
+}
+
+.home .site-header__logo .logo-ceph__text path {
+  fill: #e9eff6; /* TBC */
```
adamduncan, line 43: Maybe a `transition: fill var(--animation-easing) var(--animation-duration-fast)` here so the logo colour change is as silky smooth as the rest? 😮

src/assets/svgs/icon-close.svg
Pete-Robelou: I've omitted `<title>` from not meaningful SVGs, would that be enough?

src/_includes/components/navigation-secondary.njk
Pete-Robelou: I would prefer a solution like this but I can't get it working.

When logging out the primaryNav there is always children, I don't think it's taking into consideration the home/root of the site.

src/css/component.nav-primary.css
Pete-Robelou: Good spot ;)

src/css/component.site-header.css
Pete-Robelou: I think we will need an `overflow-y: auto` as based on the latest design of the language picker it could add extra height which could push it beyond the device viewport height, so yeah we'd need to scroll.

src/css/component.site-header.css
Pete-Robelou: Agreed.

Also added to the `close` button too.

src/_includes/components/navigation-secondary.njk
Pete-Robelou: Let's revisit this as  it's a bit of an iceberg.

src/assets/svgs/icon-close.svg
Pete-Robelou: Added `focusable="false"` and `aria-hidden="true"`.

src/js/off-canvas.js
Pete-Robelou: Agreed.

Using data attributes now.

src/js/off-canvas.js
Pete-Robelou: Cool, let's revisit this in a refactor JS ticket.

PR #14, opened 2021-01-25
Feature/2 language switcher

Pete-Robelou:
Language switcher created with details/summary components so no JS needed.

There's also a polyfill for IE.


adamduncan (approved):
Looks great, @Pete-Robelou 👏

PR #65, opened 2021-02-03
Feature/75 matrix modules

Pete-Robelou:
I've created a `matrix` CSS component which works in conjunction with a bunch of utility classes to handle this.

It's possible we may want to look at wrapping up the utility classes into part of the `matrix` CSS but let's assess this further into the build.

### DIFF
--- a/src/css/component.matrix.css
+++ b/src/css/component.matrix.css
@@ -1,0 +1,118 @@
+/* ---------- COMPONENT matrix ---------- */
+
+.matrix {
+  --matrix-gap-x: var(--size-4);
+  --matrix-gutters-x: var(--size-3);
+  --matrix-item-size: var(--size-72);
+
+  @media (--mq-sm) {
+    --matrix-gap-x: var(--size-7);
+    --matrix-gutters-x: var(--size-5);
+    --matrix-item-size: var(--size-80);
+  }
+
+  @media (--mq-md) {
+    --matrix-gutters-x: var(--size-7);
+    --matrix-item-size: var(--size-96);
+  }
+
+  @media (--mq-lg) {
+    --matrix-item-size: var(--size-72);
+  }
+
+  @media (--mq-xl) {
+    --matrix-item-size: var(--size-80);
+  }
+}
+
+.matrix__list {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.matrix__link {
+  color: var(--color-white);
+  text-decoration: none;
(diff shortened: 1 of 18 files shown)


src/css/component.matrix.css
```diff
@@ -0,0 +1,118 @@
...
+  text-decoration: underline;
+}
+
+@media (--mq-to-lg) {
+  .matrix {
+    display: flex;
+    margin-left: calc(-1 * var(--matrix-gutters-x));
+    margin-right: calc(-1 * var(--matrix-gutters-x));
+    overflow-x: auto;
+    scroll-snap-type: x mandatory;
```
adamduncan, line 51: Tasty progressive enhancement 🤙

src/css/component.matrix.css
```diff
@@ -0,0 +1,118 @@
...
+  .matrix__list > li:nth-last-child(4):first-child {
+    grid-column: 1 / span 2;
+    grid-row: 1 / span 2;
+  }
+
+  .matrix__list > li:nth-last-child(4):first-child ~ :nth-child(2) {
+    grid-column: 3 / span 2;
+  }
+
+  .matrix__list > li:nth-last-child(4):first-child ~ :nth-child(3) {
```
adamduncan, line 94: Just a note: Might be one to document to ensure that content editors understand the variants that exist, and how it behaves with differening amounts of items. (Are we shooting to support variations with just 3 and 4 items for now?)

src/css/utilities.typography.css
```diff
@@ -56,6 +56,14 @@
   line-height: var(--font-leading-normal);
 }
 
+.text-leading-8 {
```
adamduncan, line 59: Curious about these leading-8/10s. Is this mixing the concept of _size_ and _type_? What scenarios might we need this over our baseline definition of line-heights?

adamduncan (approved):
Looking good, @Pete-Robelou 👍

src/css/component.matrix.css
Pete-Robelou: Yeah, I've based this on the designs. I only saw a 3 and 4 layout.

It's a relatively easy fix to add CSS for more or less.

I can add comments to the README's for pages that use matrix modules defining the use of 3 or 4 modules.

src/css/utilities.typography.css
Pete-Robelou: Ah, these were supposed to be removed as they're not needed.

PR #67, opened 2021-02-04
CSS update to section--full allowing full screen width when inside wr…

Pete-Robelou:
…apper.


src/css/object.section.css
```diff
@@ -0,0 +1,30 @@
...
+  @media (--mq-md) {
+    --section-gutters-y: var(--size-14);
+  }
+
+  @media (--mq-lg) {
+    --section-gutters-y: var(--size-16);
+  }
+}
+
+.section--full {
```
adamduncan, line 22: Thought: Curious about the `section`>`wrapper` vs. `wrapper`>`section` approaches on the table.

I guess with this approach of having a `wrapper` around the whole page (i.e. first child of `main`) we'll get forgo the need to nest a `wrapper` in every `section`. Then in which case it's just the `section--full`s that will likely need the inner `wrapper` to suck them back in after they've been pulled out to the viewport width?

src/en-GB/discover/technology/index.html
```diff
@@ -6,32 +6,34 @@
...
-  </ul>
+    <ul class="grid lg:grid--cols-3 list-none mb-0 p-0 ul">
+      <li class="bg-white border-grey-500 border-px border-solid p-5 relative rounded-2">
+        <img class="mb-4 w-12" src="/assets/svgs/icon-benefits-blue-red.svg" alt="" />
+        <h3 class="h3">Technology</h3>
+        <p class="p">
+          Donec dignissim vestibulum lorem, sed pulvinar erat consequat sed. Quisque elementum sem eget magna feugiat mattis.
+        </p>
+        <a class="a link-cover link-cover--shadow" href="/{{ locale }}/discover/technology/">Learn more about the technology behind Ceph</a>
+      </li>
```
adamduncan, line 20: These are looking nice. Looks like we want to chat about componentisation fairly soon now modules like these are defined?

adamduncan (approved):
Nice adjustment 👍

src/css/object.section.css
Pete-Robelou: Exactly, it was simply a numbers game 🙂 

There are way more `section`s that won't be full width so a global `wrapper` made more sense. For the few `section--full`s we'll give them the inner `wrapper`.

src/en-GB/discover/technology/index.html
Pete-Robelou: Agreed.

When I start page building I'll add a few `<!-- start -->....<!-- end -->` around potential components to help rationalise them. We can also make reference to this in the README's we'll create for administering the pages.

PR #68, opened 2021-02-04
Feature/91 billboard

Pete-Robelou:
- Billboard layout with utility classes
- Moved navigation-secondary from base to individual layouts
- Frontmatter to add class to navigation-secondary
- Support layout


src/_includes/layouts/listing-blog-post-categories.njk
```diff
@@ -38,6 +38,8 @@
 }
 ---
 
+{% include 'components/navigation-secondary.njk' %}
```
adamduncan, line 41: Pondering: If this is so prevalent, I do wonder whether maybe in future we would create a nested "_subnav" layout that these could all inherit from instead of `base`. Not a big shift, but perhaps just DRYs out the includes throughout. Food for though. Maybe that's an unnecessary layer of misdirection.

adamduncan (approved):
Nice billboards 👍

src/_includes/layouts/listing-blog-post-categories.njk
Pete-Robelou: I'm on the fence with a nested layout.

Let's progress with some page building and see how we feel.

PR #69, opened 2021-02-05
Feature/84 off canvas issues

Pete-Robelou:
- Position: fixed solution to prevent `site-header` from disappearing if you opened the off-canvas once you had scrolled down the page
- Additional div to handle scrolling the navigation items within the off-canvas
- JS media query syntax fix


src/css/settings.sizes.css
```diff
@@ -64,5 +64,5 @@
   --size-h-screen: 100vh;
   --size-w-screen: 100vw;
 
-  --size-scroll-top: 5.25rem; /* 84px */
+  --size-scroll-offset: 5.25rem; /* 84px */
```
adamduncan, line 67: Would perhaps add a comment here `/* 84px — Header height */`? It's always going to be tricky to know that this value is important and if it changes, header should change. I know we all hate magic numbers, but t'is what it is.

Or sometimes I like a cheeky custom prop to give us the chance to inject a bit of meaning:

```css
--header-height: 5.25rem; /* 84px */
--size-scroll-offset: var(--header-height);
```

src/js/nav-secondary.js
```diff
@@ -7,7 +7,7 @@
   if (navigation) {
     const toggle = navigation.querySelector('[data-nav-secondary-toggle]');
     const navigation_sub = navigation.querySelector('[data-nav-secondary-sub]');
-    const media_query = window.matchMedia('(width < 64em)');
+    const media_query = window.matchMedia('(max-width: 64em)');
```
adamduncan, line 10: Nice 👍 

Nitpick: technicallllly, as we're using `max-height` should this be 1023px (63.9375em). I.e. this behaviour goes away at 1024 (64em) 🤔

adamduncan (approved):
Nice fixes @Pete-Robelou. Just one minor layout thing, but good to go otherwise 👍

src/_includes/components/site-header.njk
```diff
@@ -1,5 +1,6 @@
-<header class="site-header" data-site-header>
-  <div class="flex flex--align-center flex--justify-between wrapper">
+<div class="scroll-offset" data-scroll-offset></div>
+<header class="site-header wrapper" data-site-header>
```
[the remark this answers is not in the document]
Pete-Robelou: It will all be fine once this branch is merged with `develop` as that already contains the fixes for the full width `site-header`.

src/css/settings.colors.css
```diff
@@ -98,7 +98,20 @@
   --color-youtube-red-hsl: var(--color-youtube-red-h), var(--color-youtube-red-s), var(--color-youtube-red-l);
   --color-youtube-red: hsl(var(--color-youtube-red-hsl));
 
-  /* ---------- shadow ---------- */
+  /* ceph */
```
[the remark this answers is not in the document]
Pete-Robelou: These come from the actual Ceph logo on the ceph.io site. They're different from the `grey-300` and `grey-500` values as they were chosen by Callum.

I need them for the logo in the site-header as it changes between a light and dark version and they need to be official Ceph colours.

src/css/settings.sizes.css
Pete-Robelou: Yeah, fair point.

I've added the custom prop.

src/js/nav-secondary.js
Pete-Robelou: Aha, good thinking.

I'll add the same to the `off-canvas.js` too.

PR #74, opened 2021-02-16
Press release listing and card layout, tweaks to article card, CSS up…

Pete-Robelou:
…dates.

### DIFF
--- a/src/_includes/layouts/listing-press-releases.njk
+++ b/src/_includes/layouts/listing-press-releases.njk
@@ -8,33 +8,36 @@ permalink: '/{{ locale }}/news/press-releases{% if pagination.pageNumber > 0 %}/
 {% set pressReleaseLocaleTag = locale + '-press-release' %}
 {% set articleLocaleTag = locale + '-article' %}
 {% set pressReleaseFilters = collections[pressReleaseLocaleTag] | collectionTags([pressReleaseLocaleTag, articleLocaleTag]) %}
+{% set pressReleases = pagination.items %}
 
 {% include 'components/navigation-secondary.njk' %}
 
-<div class="flex flex--justify-between">
-  <h1>{{ title }}</h1>
-  <div>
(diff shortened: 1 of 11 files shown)


src/_includes/layouts/listing-press-releases.njk
```diff
@@ -8,33 +8,51 @@ permalink: '/{{ locale }}/news/press-releases{% if pagination.pageNumber > 0 %}/
...
-          {{ item }}</a>
-        </option>
-      {% endfor %}
-    </select>
+<section class="section">
+  <div class="flex flex--gap-6 flex--align-center flex--justify-between flex--wrap mb-16 xl:mb-24">
+    <h1 class="h1 mb-0">{{ title }}</h1>
+    <div>
+      <select class="select" id="filters" onchange="javascript:location.href = this.value;">
+        <option value="" selected disabled hidden>Category</option>
```
adamduncan, line 20: Curious if `hidden` is what we want here. 

Use case: Go to Press releases listing, select a category (and navigated to that category listing). Hit back in your browser. The state of the selected category will persist here in browsers, yet I then can't navigate to that same category again should it already be selected. I.e. can't trigger `onchange` event as there are no other optoins visible.

Screenshot 2021-02-16 at 18 58 48

Wonder if we might instead need to do conditional `selected` attributes on the `<option>` elements (i.e. if page slug === `value`) so we can also render this control on the category pages?

adamduncan (approved):
👍 

Might need to sync up on the point re: category `select` separately

src/_includes/layouts/listing-press-releases.njk
Pete-Robelou: I've removed `hidden` and `disabled` as they were causing issues.

The listing page now has the default `category` selected and the individual category pages have the chosen `category` selected.

Created a new `endsWith` filter to assist with the `if` statement to check the chosen category against the page.url.

PR #80, opened 2021-02-22
Discover layout, additional utility CSS

Pete-Robelou:


adamduncan (approved):
Looks great, @Pete-Robelou.

As you mentioned, some discussions to be had with the team around content lenght, but good to get it all in so we can see what we're working with. When we do, we can also have some conversations regarding how content is managed on these modular hub pages.

src/en-GB/discover/index.html
```diff
@@ -7,15 +7,193 @@
...
+    </div>
+    <div>
+      <h3 class="h3">Any use case</h3>
+      <p class="p">
+        Organisations that use and store data of any type can use Ceph. From start-up enterprises to academic institutions, the benefits of
+        Ceph ensure intelligent data management and insights, leading to better business decision. For those looking to secure and protect
+        their data, Ceph provides reassurance organisations need, including automatic data redundancy, no single point of failure and
+        self-managing clusters.
+      </p>
+      <a class="a inline-block mb-8 lg:mb-12" href="/{{ locale }}/solutions/case-studies/">Find out how Ceph is used</a>
```
adamduncan, line 182:
```suggestion
      <a class="a inline-block mb-8 lg:mb-12" href="/{{ locale }}/solutions/">Find out how Ceph is used</a>
```

I'd suggest we just link this straight to the Solutions hub.

The `/solutions/case-studies` page is more of a "it exists because it's needed to create a sensible hierarchy" (and to set a precedent for the filtered Case study category pages), but not something that was intended as a place to drive users to, based on the initial wires.

PR #97, opened 2021-03-03
Tech talks layout, dictionary items, table CSS update.

Pete-Robelou:
- Community tech talks layout
- Dictionary items
- Table CSS update

### DIFF
--- a/src/en-GB/community/tech-talks/index.html
+++ b/src/en-GB/community/tech-talks/index.html
@@ -3,52 +3,89 @@ title: Tech talks
 order: 2
 ---
 
-<div class="grid grid--cols-2 bg-grey-300">
-  <div>
-    <h1>{{ title }}</h1>
-    <p class="p">
-      Every month the Ceph community convenes to discuss a technical topic related to Ceph. Our hope is that the community can help to raise
-      the general level of awareness and technical knowledge of all Ceph users and contributors. Please join us in the virtual meeting to
-      ask questions and participate!
-    </p>
-    <ul class="ul">
-      <li><strong>When</strong>: 4th Thursday of every month at 17:00 UTC</li>
-      <li><strong>Where</strong>: Blue Jeans teleconference</li>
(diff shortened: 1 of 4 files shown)


src/en-GB/community/tech-talks/index.html
```diff
@@ -3,52 +3,89 @@
...
+        <tbody>
+          {% for item in item.results %}
+          <tr>
+            <td class="w-36">
+              <time datetime="{{ item.date }}">{{ item.date | formatDate(locale) }}</time>
+            </td>
+            <td>
+              {% if item.url %}
+              <a class="a" href="{{ item.url }}" target="_blank" rel="noreferrer noopener">{% endif %}{{ item.title }}{% if item.url %}</a
+              >{% endif %} {% if item.presenter %}<span class="inline-block text-lower">{{ 'with' | i18n }} {{ item.presenter }}</span>{%
```
adamduncan, line 79: @Pete-Robelou Is there a reason we `text-lower`'d this? The `tech_talks` content (data) is properly cased with names etc. Perhaps we render as is?

Pete-Robelou: It was to lowercase the word `with` as grammatically the whole line reads as a sentence. I guess we could just wrap a `span` around the word `with` so only that is lowercase ?

Screenshot 2021-06-23 at 11 07 54
Screenshot 2021-06-23 at 11 08 07

adamduncan: Ah, gotcha. Makes sense. Yeah, good shout. Let's just apply that to `with` 👍

Pete-Robelou: I've removed the `<span class="inline-block">` as this was originally to improve the wrapping on smaller screens, however in retrospect it's not really an improvement 🙂 . 

We're now only wrapping `with` to make it lowercase.

PR #99, opened 2021-03-09
Feature/42 blog listing

Pete-Robelou:
- Rollup config update to support to dynamic require targets and mixed es modules
- Restructured the search results data to match articleCard requirements
- Removed url from search raw and index


rollup.config.js
```diff
@@ -8,13 +8,27 @@ const distDir = 'dist/js/';
 
 const pluginsES = () => [
   resolve(),
-  commonjs(),
-  process.env.NODE_ENV === 'production' ? terser({ output: { comments: false } }) : null,
+  commonjs({
+    dynamicRequireTargets: [
+      './src/_11ty/**/*.js',
```
adamduncan, line 13: Just a thought: I wonder if we go a bit tighter on what gets declared here. More of an narrow/opt-in scope (i.e. this extra `dynamicRequireTargets` and `transformMixedEsModules` just to serve the 11ty shortcode ArticleCard module (and by extension the filters it relies on), because it's CommonJS?)

```js
dynamicRequireTargets: [
  './src/_11ty/filters/**/*.js',
  './src/_11ty/shortcodes/**/*.js',
],
```

There's an ongoing [discussion](https://github.com/11ty/eleventy/issues/836) about native Node support for ES modules, so hopefully this is just a temporary stopgap until we can move our 11ty scripts to the new format and drop any `module.exports`/`require` malarkey.

src/_includes/layouts/listing-blog-post-categories.njk
```diff
@@ -38,15 +38,69 @@
...
 
-<ul class="ul grid grid--cols-4">
-  {% for item in blogPostCategories %}
-    <li>
-      {% include "components/blog-post-card.njk" %}
-    </li>
-  {% endfor %}
-</ul>
\ No newline at end of file
+<section class="section">
```
adamduncan, line 46: Thought: Feels like the bulk of the templated part of this layout is the same whether a blog-category/blog-list page. Opportunity to make a generic blog list template that both could inherit from? (With just a smidge of conditional logic to omit the "Articles filtered by..." title). 

Maybe a premature optimisation, but feels some could be DRYed out here

src/_includes/layouts/listing-blog-search.njk
```diff
@@ -3,15 +3,64 @@ layout: navigation
...
+            <path d="M16.294 14.579h-.903l-.32-.31a7.4 7.4 0 001.795-4.836 7.433 7.433 0 10-7.433 7.433 7.4 7.4 0 004.837-1.796l.309.32v.904L20.296 22 22 20.296l-5.706-5.717zm-6.861 0a5.139 5.139 0 01-5.146-5.146 5.139 5.139 0 015.146-5.146 5.139 5.139 0 015.146 5.146 5.139 5.139 0 01-5.146 5.146z"/>
+          </svg>
+          <span class="visually-hidden">Submit search</span>
+        </button>
+      </form>
+    </div>
+  </div>
 
-<div id="search-results"></div>
+  <div id="search-results"></div>
```
adamduncan, line 45: Suggestion: Perhaps worth a `min-height` on the results container here for less of a page-jump.

We could even take it a step further in the `search` script, and toggle an `is-loading` class, perhaps while the search initialises? Might also give us an avenue to  a `<noscript>` version.

(Food for thought. All likely future enhancements)

src/css/base.forms.css
```diff
@@ -1,6 +1,75 @@
 /* ---------- BASE forms ---------- */
 
-/* ---------- select ---------- */
+/* input-search */
+.input[type="search"] {
```
adamduncan, line 4: Question: Are there any other `input type="text"`s in the site? Wondering if this becomes more of a generic catch all for most "traditional" `input` types (text, date, email, phone, etc.) as well as search

src/_11ty/shortcodes/ArticleCard.js
```diff
@@ -0,0 +1,60 @@
...
+const removeHtml = require(`${filtersDir}/removeHtml.js`);
+const truncate = require(`${filtersDir}/truncate.js`);
+
+module.exports = (
+  {
+    data: { author = '', date, image, tags, title = '', locale = '' },
+    templateContent,
+    url,
+  },
+  label = ''
```
adamduncan, line 13: Suggestion: Perhaps if we think we might need any additional configuration for `ArticleCard`, we might consider its API: `ArticleCard(item, options)`

Just thinking, it could be more self-describing, and leave us open to other options in future w/o having to rework the existing implementations. E.g.

```js
ArticleCard(item, {
  showLabel: true
})
```

adamduncan (approved):
Looking good @Pete-Robelou 👍 

Only a few suggestions really. Non-blocking, just some ideas to chew on.

src/_includes/layouts/listing-blog-post-categories.njk
```diff
@@ -38,15 +38,69 @@
...
+  <div class="grid grid--align-center lg:grid--cols-auto-160 mb-16 xl:mb-24">
+    <h1 class="h1 mb-0">{{ title }}</h1>
+    <div class="grid md:grid--cols-2">
+      <div>
+      <select class="select w-full" id="filters" onchange="javascript:location.href = this.value;">
+        <option value="">{{ 'category' | i18n }}</option>
```
adamduncan, line 52:
```suggestion
        <option value="/{{ locale }}/news/blog">{{ 'category' | i18n }}</option>
```

Could we add the blog's root URL here, so we can effectively "remove" the category filter and move back to the full listing via this control?

src/_includes/layouts/listing-blog-post-categories.njk
Pete-Robelou: @adamduncan The big difference between those layouts is the pagination. 

`listing-blog-posts` uses pagination in the traditional sense that it creates multiple pages of posts where as `listing-blog-categories` uses it to automate the creation of a page for each individual category based on a tag.

Maybe we can look at this as a future optimisation but I think for now we should leave as is.

src/css/base.forms.css
Pete-Robelou: @adamduncan At this stage the only `input` is the search on the blog page.

If we can additional inputs in the future it'll be an easy CSS update.

src/_includes/layouts/listing-blog-search.njk
Pete-Robelou: @adamduncan Adding a `min-height` is a good idea.

Like you say I think an `is-loading` would be a nice future enhancement.

src/_11ty/shortcodes/ArticleCard.js
Pete-Robelou: @adamduncan Good idea.

I'll update that.

src/_includes/layouts/listing-blog-post-categories.njk
Pete-Robelou: @adamduncan Yeah, I think that would work.

rollup.config.js
Pete-Robelou: Changed.

PR #107, opened 2021-03-11
ansible: Also include the default mime.types

djgalloway:
I guess specifying `types { }` in a block higher than `http { }` overrides all of the default mime types so we need to manually include it at the same level again.

https://stackoverflow.com/a/20566966/6225562


adamduncan (approved):

adamduncan:
Thanks @djgalloway

PR #115, opened 2021-04-21
index: adjust prose a bit

liewegas:


adamduncan:
Nice one, @liewegas. Happy for this one to go in 👍 

What's your preferred workflow for content contributions like this? Perhaps one to sync on as a team in our next session?

Issue #104, opened 2021-03-10
Gzip static resources


adamduncan:
Keen to get some thoughts from the server-side.

Is it possible to serve our static resources (especially `text/css` and `text/javascript` with Gzip compression?

We could see some really effective performance improvements on this front: https://www.webpagetest.org/performance_optimization.php?test=210310_Ai74_0b3ca30afa6828badd8994e5f6bc1e31&run=3#compress_text

Is that easier said than done? 😄

djgalloway:
https://github.com/ceph/ceph.io/pull/106/commits/56715ec713ff916de4abaf71b1acc3cc2b8f2ddb

adamduncan:
Confirm static CSS + JS assets being gzipped at develop.ceph.io. Thank you

PR #130, opened 2021-05-26
Add `application/json` to gzip types, `mjs` MIME type update

adamduncan:
Following on from #104, it would be great if we could gzip `application/json` files. We'll have a few static JSON assets (search indices, etc.) which would benefit a lot from compression.

@djgalloway is this fair game/all we need to include?

Related, we currently have `text/javascript` included in our gzip types, but we set the `.mjs` MIME type to `application/javascript`. Based on [a little](https://github.com/google/WebFundamentals/issues/7549) [digging](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types), I've also updated this to `text/javascript` to ensure they're gzipped as intended. Shout if this isn't in line with your expectations.

### DIFF
--- a/ansible/roles/deploy/templates/site.j2
+++ b/ansible/roles/deploy/templates/site.j2
@@ -55,7 +55,7 @@ server {
   # https://github.com/ceph/ceph.io/issues/101
   include mime.types;
   types {
-    application/javascript mjs;
+    text/javascript mjs;
   }
 
   # https://github.com/ceph/ceph.io/issues/104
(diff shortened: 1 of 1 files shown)


djgalloway:
From [this comment](https://github.com/google/WebFundamentals/issues/7549#issue-437241005):

> I was told that `text/javascript` is obsolete and it should be `application/javascript`

I feel like I remember that from adjusting this setting on another one of my public sites.  docs.ceph.com maybe?  Anyway, I just checked out this branch and pushed the changes to nginx.  If it looks right to you (i.e., no errors), I'm fine with it.

adamduncan:
>  If it looks right to you (i.e., no errors), I'm fine with it.

Cool, thanks for following up @djgalloway. Yep, sounds like there's been some confusion around this MIME type, but happy to go with the [spec](https://html.spec.whatwg.org/multipage/scripting.html#scriptingLanguages:javascript-mime-type) (and can't see it throwing up any issues with JS sources). Thanks.

PR #131, opened 2021-05-27
Search JSON cleaning and error handling

### LINKED ISSUE #118: Search Index: Filtering out characters
To prevent the search index from falling over, which can happen if there is an incompatible character in the article content, we should implement a stricter approach in our `cleanIndex` filter and remove more characters.

It's probably best to go wide with this approach as we won't know for sure what content will be in an article.

Pete-Robelou:
- Renamed cleanIndex to cleanSearchRaw
- Revised filtering to remove more stuff including unicode characters
- New cleanSearchOutput filter to optimise what we see in the search results cards
- Installed npm chalk package
- try/catch script to prevent build from running

Closes #118


src/_11ty/filters/cleanSearchRaw.js
```diff
@@ -0,0 +1,42 @@
...
+
+  // replace line breaks
+  var lineBreaks = /[\r\n]+/gm;
+  result = result.replace(lineBreaks, ' ');
+
+  // replace extra spaces
+  var extraSpaces = /\s+/g;
+  result = result.replace(extraSpaces, ' ');
+
+  return result;
```
adamduncan, line 41: Personal: I'd perhaps look at chaining the replaces on the string, as to keep the code terse and clear separation of definition (i.e. regexes) and execition (i.e. replaces). Saves on reassigning the `result` with each change, too. E.g.

```js
const shortWords = /\b(the|a|an|and|am|you|I|to|if|of|off|me|my|on|in|it|is|at|as|we|do|be|has|but|was|so|no|not|or|up|for|ve|ll|re|s)\b/gi;
const unicode = /[\u0000-\u001F\u007F-\u009F]/g;
const punctuation = /[!“”‘’"#$%&'()*+,-./\\:;<=>?@[\]^_`{|}~]/g;
const spaces = /[\r\n\s]+/gm;

return dedupedStr.replace(shortWords, '')
         .replace(unicode, '')
         .replace(punctuation, '')
         .replace(spaces, ' ')
```

adamduncan (approved):
Implementation looks like that should do the trick. Haven't tried this against the blog import PR, but last we tested, the `unicode` seemed to be the kicker to resolve any build issues. 👍

src/_11ty/filters/cleanSearchRaw.js
Pete-Robelou: Fair point.

I've updated both `cleanSearchOutput.js` and `cleanSearchRaw.js`.

PR #132, opened 2021-05-28
Fall back images

### LINKED ISSUE #117: Articles: Fallback images
A lot of legacy articles won't have hero images so when they appear on the blog listing page their card will be missing an image. Our current solution checks if an image exists in the frontmatter, if not it is replaced with a single fallback image. This solution means that all articles without a hero image share the same fallback image. 

We should look into a solution that creates a score for each article, perhaps based on the article title, and matches that score against a pool of generic fallback images. This is better than randomly picking images as it means the articles will consistently show the same fallback image everytime the site runs a new build. We need to make sure we have a large enough pool of fallback images to avoid repetition.

Pete-Robelou:
- getSingleDigitFromDate filter
- fall back image assets
- filter updates that used site.defaultLocale

I've only applied the new filter for fallback images to the `ArticleCard` used for `blog posts` and `press releases`.

It's feasible that `events` and `case studies` may need also need fallback images but it feels more achievable to source proper images for these. We would also need to re-consider the current solution as it relies on a `date` to generate a random number and `case studes` don't have a date.

Closes #117


src/_11ty/filters/formatDate.js
```diff
@@ -1,4 +1,7 @@
-module.exports = (date, locale = site.defaultLocale) => {
+const site = require('../../_data/site');
+const { defaultLocale } = site;
+
+module.exports = (date, locale = defaultLocale) => {
```
adamduncan, line 4: Good catch. Didn't realise global data isn't available to filters. Perhaps the v1 of 11ty might have neater ways of getting access to that, or we can look at passing it along from the 11ty config, so these filters remain pure.

adamduncan (approved):
Nice solution, @Pete-Robelou 👍

src/_11ty/filters/formatDate.js
Pete-Robelou: @adamduncan I done a little research and couldn't find anything relating to making global data available to filters.

It does seem like [v1 will allow global data to be added to the 11ty config](https://www.11ty.dev/docs/data-global-custom/) but I still can't find any mention of this being available to filters.

Let's revisit this once/if we upgrade to v1.

Pete-Robelou: 🤔 I guess it maybe possible to pass along the `defaultLocale` if we added the filter directly to the config file rather than requiring it.

So this syntax:
```
eleventyConfig.addFilter("formatDate", function(value) { … });
```

rather than
```
eleventyConfig.addFilter('formatDate', require(`${filtersDir}/formatDate.js`));
```

PR #134, opened 2021-05-28
Richtext image width

### LINKED ISSUE #120: Richtext CSS: Image width
Investigate a solution so inline richtext images remain responsive but also don't expand wider than their native width.

Our current implemention is a bit opinionated and forces all images to be 100% width of their parent container.

Pete-Robelou:
- Richtext CSS so images are no wider than their native width but remain responsive

We have a global declaration for `img` that forces images to be 100% width of their parent container. I've removed this so where necessary we'll have to use the `w-full` utility class if we want an image to be 100% width of its container. It's been removed because as long as we have a global width declaration in the CSS, whether it be `auto` or `100%`, we won't be able to honor a width attribute on an ``. It can be overriden by adding a `class` to the `img` but it's in the interest of keeping the markdown simple and not needing an awareness of our utility width classes.

Images that do appear in a `richtext` container will now be no wider than their native width but remain responsive, they will also appear center aligned.

If an alternative width is required for `richtext` images we'd need to use the proper HTML markup `` with a width declaration. The image would still be responsive but no wider than the declared width.

Closes #120

### DIFF
--- a/src/_includes/components/case-study-category-card.njk
+++ b/src/_includes/components/case-study-category-card.njk
@@ -1,8 +1,10 @@
 {# We will need a way to randomise the image as case study categories don't contain an image. 
 We should only need a selection of 6 assets as we only show 6 case study categories. #}
 
+{# UPDATE - This component may no longer be needed due to the IA update moving case studies under discover #}
+
 <div class="bg-grey-300 border-grey-500 border-px border-solid relative rounded-2">
-  <div class="aspect-ratio aspect-ratio--16x9">
+  <div class="aspect-ratio aspect-ratio--16x9 aspect-ratio--cover">
     <img alt="" class="absolute h-full left-0 rounded-t-2 top-0 w-full" loading="lazy" src="/assets/bitmaps/photo-water-02.jpg" />
   </div>
   <a class="block color-red-500 h3 link-cover link-cover--shadow mb-0 p-6 text-capitalise" href="/{{ locale }}/discover/case-studies/category/{{ item | slug }}/">{{ item }}</a>
--- a/src/_includes/components/case-study-category-card.njk
+++ b/src/_includes/components/case-study-category-card.njk
@@ -3,7 +3,7 @@ We should only need a selection of 6 assets as we only show 6 case study categor
 
 <div class="bg-grey-300 border-grey-500 border-px border-solid relative rounded-2">
   <div class="aspect-ratio aspect-ratio--16x9">
-    <img alt="" class="absolute left-0 rounded-t-2 top-0" loading="lazy" src="/assets/bitmaps/photo-water-02.jpg" />
+    <img alt="" class="absolute h-full left-0 rounded-t-2 top-0 w-full" loading="lazy" src="/assets/bitmaps/photo-water-02.jpg" />
   </div>
   <a class="block color-red-500 h3 link-cover link-cover--shadow mb-0 p-6 text-capitalise" href="/{{ locale }}/discover/case-studies/category/{{ item | slug }}/">{{ item }}</a>
 </div>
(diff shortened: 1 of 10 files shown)


src/_includes/components/case-study-category-card.njk
```diff
@@ -3,7 +3,7 @@ We should only need a selection of 6 assets as we only show 6 case study categor
 
 <div class="bg-grey-300 border-grey-500 border-px border-solid relative rounded-2">
   <div class="aspect-ratio aspect-ratio--16x9">
-    <img alt="" class="absolute left-0 rounded-t-2 top-0" loading="lazy" src="/assets/bitmaps/photo-water-02.jpg" />
+    <img alt="" class="absolute h-full left-0 rounded-t-2 top-0 w-full" loading="lazy" src="/assets/bitmaps/photo-water-02.jpg" />
```
adamduncan, line 6: Question: Do we want these to be `height: auto`, should the images being uploaded not always be cut to perfect 16:9 ratio (as might happen over time)? Or do we want to use `h-full` and also ensure it has `object-fit: cover`?

src/_includes/layouts/blog-post.njk
```diff
@@ -21,7 +21,7 @@ layout: navigation
       {% if image %}
         <div class="to-lg:w-full-breakout">
           <div class="aspect-ratio aspect-ratio--16x9 aspect-ratio--contain bg-grey-500 mb-8 lg:mb-10 xl:mb-12">
```
adamduncan, line 23: We'll perhaps have to keep an eye on how the blog post image creation goes here. Not sure I'm 100% on enforcing 16:9 w/ contain will result in the desired effect?

Would we  look at doing away with the fixed-ratio container in some circumstances, and instead let the browser try to hold space in the page by giving images `width`/`height` attributes?

Side note: We might also be able to get that dimensions metadata should we explore something like the [11ty Image plugin](https://www.11ty.dev/docs/plugins/image/) in future 👀

adamduncan (approved):
👍 Looks like a nice improvement. Suspect we'll want to keep a cheeky eye on this as we get a better idea of how asset production/population is going

src/_includes/components/case-study-category-card.njk
Pete-Robelou: @adamduncan In the context of a `card` I think we should go with `h-full` and `w-full` and probably use `object-fit: cover` otherwise we can end up in a situation where the image won't fill all the space requiring us to use a bg color.

E.g. This is actually the current implementation for the `articleCard` which uses `object-fit: contain` so it doesn't crop any of the image.

Screenshot 2021-06-03 at 15 37 45

I agree that it's very likely over time the images will not be a perfect 16:9 ratio but when it comes to `cards` we need to be consistent with the layout so using `h-full`, `w-full` and `object-fit: cover` will be the best approach.

If there's no objections I'll go back through the code and update to `aspect-ratio--cover` for all the cards.

src/_includes/layouts/blog-post.njk
Pete-Robelou: The 11ty Image plugin looks nice and definitely worth considering. Another option is the [CSS aspect-ratio](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio) attribute, although it doesn't have support in Safari (at the moment ) or IE. It has an `auto` option which should work for images that don't adhere to a proper ratio.

One to revisit in the future.

src/_includes/components/case-study-category-card.njk
adamduncan: Yep, sounds great mate. Thanks 👍

src/_includes/layouts/blog-post.njk
Pete-Robelou: I've done a little R&D into the 11ty Image plugin but this can be continued on it's own branch as it raises some new questions.

I've removed the `aspect-ratio` from the hero images on the editorial content pages like `blog-post`, `press-releases` etc...

PR #135, opened 2021-05-28
YouTube Shortcode

### LINKED ISSUE #119: YouTube: Shortcode
To facilitate simple content editing for all the markdown articles we should create a `YouTube` shortcode.

This will sweep away the markup so it doesn't complicate the markdown and just require the user to enter a YouTube ID.

In the markdown it would look something like this:
```
{% YouTube 'PhxVPEZeHp4' %}
```

Pete-Robelou:
- YouTube shortcode with props for id and title
- wrapped with aspect-ratio to make it responsive
- .richtext .aspect-ratio CSS

Closes #119

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,3 +156,31 @@ title: Don't repeat yourself
 
 # {{ title }}
 ```
+
+### Shortcodes
+
+Shortcodes are reusable code snippets that allow us to sweep away complicated scripts into a nice, easy user interface.
+
+For example this is the script for a YouTube embed.
+
+```
+<iframe
+  width="560"
+  height="315"
+  src="https://www.youtube.com/embed/vQF17UBU4RE"
+  title="Ceph Tech Talk: Karan Singh - Scale Testing Ceph with 10Billion+ Objects 2020-10-01"
+  frameborder="0"
+  allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
+  allowfullscreen
+></iframe>
+```
+
+We can capture this script in a shortcode so we don't need to repeat it everytime we want to embed a YouTube video. It accepts two values, the `id` and the `title` of the video and uses the following syntax.
+
+```
+{% YouTube 'vQF17UBU4RE', 'Ceph Tech Talk: Karan Singh - Scale Testing Ceph with 10Billion+ Objects 2020-10-01' %}
+```
+
+The first argument is the name of the shortcode `YouTube` and then we pass it the id of the video `vQF17UBU4RE` followed by the title `Ceph Tech Talk: Karan Singh - Scale Testing Ceph with 10Billion+ Objects 2020-10-01`.
+
+We now have a single source of truth for the YouTube embed script making it easy to maintain and easy to reuse with a simple, clean interface.
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ Shortcodes are reusable code snippets that allow us to sweep away complicated ma
 ></iframe>
 ```
 
-We can capture this script in a shortcode so we don't need to repeat it everytime we want to embed a YouTube video. It accepts two values, the `id` and the `title` of the video and uses the following syntax.
+We can capture this markup in a shortcode so we don't need to repeat it everytime we want to embed a YouTube video. It accepts two values, the `id` and the `title` of the video, and uses the following syntax:
 
 ```
 {% YouTube 'vQF17UBU4RE', 'Ceph Tech Talk: Karan Singh - Scale Testing Ceph with 10Billion+ Objects 2020-10-01' %}
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ title: Don't repeat yourself
 
 Shortcodes are reusable code snippets that allow us to sweep away complicated markup into a nice, easy user interface.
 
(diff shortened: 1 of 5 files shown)


adamduncan (changes requested):
Looks good @Pete-Robelou 👍 

Suggest we maybe add a section to the main README regarding shortcode usage, as it's a new concept specific to these static sites.

As part of this PR, we might also want to add an example in situ, so others can reference. What if we swapped out the Tech talks' YouTube video embed for this shortcode?

Pete-Robelou:
> Looks good @Pete-Robelou 👍
> 
> Suggest we maybe add a section to the main README regarding shortcode usage, as it's a new concept specific to these static sites.
> 
> As part of this PR, we might also want to add an example in situ, so others can reference. What if we swapped out the Tech talks' YouTube video embed for this shortcode?

@adamduncan Good idea. 

I've swapped out the Tech Talk embed code for the YouTube shortcode and added a section to the README.

Let me know if you're happy with my explanation.

README.md
```diff
@@ -156,3 +156,31 @@ title: Don't repeat yourself
 
 # {{ title }}
 ```
+
+### Shortcodes
+
+Shortcodes are reusable code snippets that allow us to sweep away complicated scripts into a nice, easy user interface.
```
adamduncan, line 162:
```suggestion
Shortcodes are reusable code snippets that allow us to sweep away complicated markup into a nice, easy user interface.
```

README.md
```diff
@@ -156,3 +156,31 @@ title: Don't repeat yourself
...
+
+For example this is the script for a YouTube embed.
```
adamduncan, line 164:
```suggestion
#### YouTube video player embed
```

What if we have a sub-section per shortcode? As expect we might have more in future

README.md
```diff
@@ -156,3 +156,31 @@ title: Don't repeat yourself
...
+  height="315"
+  src="https://www.youtube.com/embed/vQF17UBU4RE"
+  title="Ceph Tech Talk: Karan Singh - Scale Testing Ceph with 10Billion+ Objects 2020-10-01"
+  frameborder="0"
+  allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
+  allowfullscreen
+></iframe>
+```
+
+We can capture this script in a shortcode so we don't need to repeat it everytime we want to embed a YouTube video. It accepts two values, the `id` and the `title` of the video and uses the following syntax.
```
adamduncan, line 178:
```suggestion
We can capture this markup in a shortcode so we don't need to repeat it everytime we want to embed a YouTube video. It accepts two values, the `id` and the `title` of the video, and uses the following syntax:
```

README.md
```diff
@@ -156,3 +156,31 @@ title: Don't repeat yourself
...
+
+```
+{% YouTube 'vQF17UBU4RE', 'Ceph Tech Talk: Karan Singh - Scale Testing Ceph with 10Billion+ Objects 2020-10-01' %}
+```
+
+The first argument is the name of the shortcode `YouTube` and then we pass it the id of the video `vQF17UBU4RE` followed by the title `Ceph Tech Talk: Karan Singh - Scale Testing Ceph with 10Billion+ Objects 2020-10-01`.
+
+We now have a single source of truth for the YouTube embed script making it easy to maintain and easy to reuse with a simple, clean interface.
```
adamduncan, line 186:
```suggestion
We now have a single source of truth for the YouTube embed code making it easy to maintain and easy to reuse with a simple, clean interface.
```

README.md
```diff
@@ -156,3 +156,31 @@ title: Don't repeat yourself
...
+></iframe>
+```
+
+We can capture this script in a shortcode so we don't need to repeat it everytime we want to embed a YouTube video. It accepts two values, the `id` and the `title` of the video and uses the following syntax.
+
+```
+{% YouTube 'vQF17UBU4RE', 'Ceph Tech Talk: Karan Singh - Scale Testing Ceph with 10Billion+ Objects 2020-10-01' %}
+```
+
+The first argument is the name of the shortcode `YouTube` and then we pass it the id of the video `vQF17UBU4RE` followed by the title `Ceph Tech Talk: Karan Singh - Scale Testing Ceph with 10Billion+ Objects 2020-10-01`.
```
adamduncan, line 184:
```suggestion
We call shortcodes by name with the Nunjucks block syntax: `{% YouTube %}`. The first argument we pass is the `id` of the video (e.g. `'vQF17UBU4RE'`). The second (optional) argument will set the `title` attribute of the iframe embed (e.g. `'Ceph Tech Talk: Karan Singh - Scale Testing Ceph with 10Billion+ Objects 2020-10-01'`).

_Note:_ Shortcode arguments are type-/space-sensitive, so should should include surrounding `'` quote marks.
```

Thanks @Pete-Robelou. Other comments are very minor terminology tweaks. 

This one though I thought could use some more explicit instruction. How's this read? Technically `YouTube` is more like the function/shortcode name, while the first argument is `id`, and second `title` (if we're thinking about the function signature).

adamduncan (approved):

README.md
Pete-Robelou: Yeah, good idea.

README.md
Pete-Robelou: Looks good.

Issue #122, opened 2021-05-18
Amend language site directory names, redirects


adamduncan:
The decision's been made to go with two-letter country codes for our language site sub-directories.

- [x] @Pete-Robelou, can we please rename our `en-GB` (and `jp-JP`, for the purposes of demonstration?) directories to their two-letter equivalent? `en` and `jp`, respectively. Expect we'll also need to amend any reference of `en-GB` where used for default/fallback locale, or collection paths, etc.
- [x] @djgalloway Once this has been done, can we please get your support in updating the redirects to reflect the change? If that's just a case of changing redirect references in `/ansible/roles/deploy/templates/site.j2` (?) then @Pete-Robelou, we can manage that as part of our code updates.

In future, should we need region-specific language identifiers for language sites (e.g. `pt-br`), we'll introduce them as necessary, and for those sites only.

djgalloway:
> If that's just a case of changing redirect references in `/ansible/roles/deploy/templates/site.j2` (?) then @Pete-Robelou, we can manage that as part of our code updates.

That's all it should take.  The playbook gets run automatically to update nginx.  I'm happy to help if that doesn't do it for some reason.

Pete-Robelou:
@adamduncan All done.

adamduncan:
Think this one's all good and deployed? Good to close?

Pete-Robelou:
Yeah, all good.

PR #140, opened 2021-06-02
Reverse order of Blog Posts and Press Releases

Pete-Robelou:
I've included the fix below as I missed it when working on #122 

- minor fix to `getArticleType` filter as we now use a 2 character prefix for language names


src/_11ty/filters/getArticleType.js
```diff
@@ -18,9 +18,9 @@ module.exports = tags => {
     'press-release': 'PRESS RELEASE',
   };
 
-  // tags to lowercase, strip out the first 6 chars and check includes tag suffix
+  // tags to lowercase, strip out the first 3 chars and check includes tag suffix
   const lcTags = tags
-    .map(tag => tag.toLowerCase().substr(6))
+    .map(tag => tag.toLowerCase().substr(3))
```
adamduncan, line 23: Mm 🤔 Let's consider this one for a second. While we've opted for the `en` convention over `en-GB` for our initial language site(s), it's not to say the requirement for a `pt-br` or similar mightn't come up in future. Is there a different way we can split the locale off the tag? Or do we instead try to leverage our known `locale` as part of the `filter()` instead? Come at it from the other angle?

Pete-Robelou: This has all been revamped and removes any issue with the locale convention `en` over `en-GB`. It's also been enhanced to support translated values whereas before it was english only.

The translated values update spawned a tangent highlighting our search results page needed some translation too so this has been included on this branch.

PR #144, opened 2021-06-04
ceph.io/roadmap: cleaning up roadmap

zdover23:
This is the first PR that removes the marketing-
speak from the Roadmap page on ceph.io.


liewegas (approved):
Just a note that we will soon be producing a nice, higher-level roadmap that will eventually appear here.  For now, trello + redmine is perfect.

PR #145, opened 2021-06-07
Language-specific 404 pages

### LINKED ISSUE #103: 404 page
To configure, based on where we permalink our language-specific 404 pages. TBC

Pete-Robelou:
- Updated primary navigation to also exclude pages tagged with 'warning'
- Basic layout of 404 pages

Closes #103


src/_11ty/collections/primary.js
```diff
@@ -3,6 +3,9 @@
 module.exports = function (collectionApi) {
   return collectionApi.getAll().filter(item => {
     const { tags = [] } = item.data;
-    return !tags.includes('support');
+
+    const excludeTags = ['support', 'warning'];
```
adamduncan, line 7: Thought: Would perhaps refer to this as `error` instead of `warning`? I.e. Could be generalised as covering [error pages](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404#custom_error_pages)

src/en/404.html
```diff
@@ -0,0 +1,18 @@
+---
+layout: base
+title: Not found
+tags: 'warning'
+permalink: '/{{ locale }}/404.html'
+---
+
+<section class="section text-center">
+  <h1 class="h1">404</h1>
+  <p class="standout">The page you are looking for is not here (anymore).</p>
```
adamduncan, line 10:
```suggestion
  <p class="standout">The page you are looking for is not here.</p>
```

Suggestion: Perhaps we remove `(anymore)` here. While it might be true of some old pages which no longer exist on the new site (those that haven't had 301 redirects put in place), it won't always be true of others, e.g. https://develop.ceph.io/bogus-page — perhaps best to just say "not here"?

adamduncan (approved):
Looks good, @Pete-Robelou.

One minor issue I ran into when running the site locally. The browsersync config contains [a small snippet](https://github.com/ceph/ceph.io/blob/develop/.eleventy.js#L122-L127) to rewrite 404s when in dev mode. Perhaps we update this to point at the new `en` 404 page, just to cover our bases?

Pete-Robelou:
> Looks good, @Pete-Robelou.
> 
> One minor issue I ran into when running the site locally. The browsersync config contains [a small snippet](https://github.com/ceph/ceph.io/blob/develop/.eleventy.js#L122-L127) to rewrite 404s when in dev mode. Perhaps we update this to point at the new `en` 404 page, just to cover our bases?

Good spot. Updated to `const content_404 = fs.readFileSync('dist/en/404.html');`.

src/_11ty/collections/primary.js
Pete-Robelou: Ha, I originally had it as `error` but then changed it to `warning`. I'll change it back to `error`.

src/en/404.html
Pete-Robelou: Yeah, that seems fair. I've updated.

Issue #136, opened 2021-06-02
Sitemap xml


Pete-Robelou:
- [x] Create sitemap xml

Some references:
- https://github.com/11ty/eleventy-base-blog/blob/master/sitemap.xml.njk
- https://www.npmjs.com/package/@quasibit/eleventy-plugin-sitemap
- https://www.11ty.recipes/recipes/add-a-sitemap/

Pete-Robelou:
@adamduncan I've got the sitemap.xml working now with a unique `translationKey` per page using `eleventyComputed.js`.

I've spotted an issue relating to pages created with `pagination`, they all use the same `href` even though they have different `loc`. I've had a look round and it seems to be a common issue that's not been resolved. https://github.com/11ty/eleventy/issues/248#issuecomment-653754037

It raises the question of whether we need to include category pages in the sitemap.xml as the pages they reference are the ones that matter, `blog-posts`, `case-studies`, `press-releases` etc... It's probably not worth including the blog search page either as on it's own it doesn't contain any content. We can remove this pages with `eleventyExcludeFromCollections: true`, what do you think ?

```
  <url>
    <loc>https://ceph.io/en/discover/case-studies/commerce/</loc>
    <lastmod>2021-06-07T09:59:39.698Z</lastmod>
    <xhtml:link rel="alternate" hreflang="en" href="https://ceph.io/en/discover/case-studies/commerce/" />
  </url>
  <url>
    <loc>https://ceph.io/en/discover/case-studies/government/</loc>
    <lastmod>2021-06-07T09:59:39.698Z</lastmod>
    <xhtml:link rel="alternate" hreflang="en" href="https://ceph.io/en/discover/case-studies/commerce/" />
  </url>
  <url>
    <loc>https://ceph.io/en/discover/case-studies/information/</loc>
    <lastmod>2021-06-07T09:59:39.698Z</lastmod>
    <xhtml:link rel="alternate" hreflang="en" href="https://ceph.io/en/discover/case-studies/commerce/" />
  </url>
```

adamduncan:
Mm, that's unfortunate, but not a show-stopper. Would be nice, I suppose, from an indexing POV for those category pages to be included. But perhaps until that issue is resolved, it's a case of excluding them from the sitemap?

Everything else looking good.

Pete-Robelou:
[...]

PR #151, opened 2021-06-08
Multilingual sitemap

Pete-Robelou:
- installed eleventy-plugin-sitemap
- sitemap collection
- translationKey with eleventyComputed data
- removed searchoutput/raw from collections
- alternate links in head for pages with language variants
- canonical link
- robots.txt

See my comments in #136 about removing some additional pages from collections.


.eleventy.js
```diff
@@ -87,6 +89,11 @@ module.exports = function (eleventyConfig) {
       '*': 'en',
     },
   });
+  eleventyConfig.addPlugin(sitemap, {
+    sitemap: {
+      hostname: 'https://ceph.io',
```
adamduncan, line 94: Got me thinking about the develop vs. prod versions of these sites. We may want some values like this to be dynamic, determined by the `NODE_ENV` the site was built for. So this would be environment-specific instead of hardcoded.

By extension we might want to consider adding `<meta name="robots" content="noindex nofollow">`, as one example, to the non-production environments 🤔 Worth another ticket?

adamduncan (approved):
This is great, @Pete-Robelou 👍  Nice work

.eleventy.js
Pete-Robelou: @adamduncan Something like this ?

```
eleventyConfig.addPlugin(sitemap, {
    sitemap: {
      hostname: process.env.NODE_ENV === 'production' ? 'https://ceph.io' : 'http://localhost:8080',
    },
  });
```

adamduncan: Potentially in that ballpark, yep. Thinking more:

```
hostname: process.env.NODE_ENV === 'production' ? 'https://ceph.io' : 'http://develop.ceph.io',
```

...to cover the develop site. Not so concerned with running locally.

Though saying that, there will be some changes we need to make to the build scripts (and CI config) to ensure we're setting `NODE_ENV` for its respective envs. At the moment `build` just assumes `production`. 

I'd say merge this feature, and let's open a dedicated ticket for dev/prod env stuff?

Pete-Robelou: Yeah, ok

PR #163, opened 2021-06-10
ceph.io/community: add tech talks (1 of several)

zdover23:
This PR adds five Ceph Month 2021 talks to the
"Tech Talks" subpage on the "Community" page,
and adds one video called "What's New in Pacific
Ceph Release". (All six things added are videos.)

### DIFF
--- a/src/en/community/community.json
+++ b/src/en/community/community.json
@@ -1,10 +1,40 @@
 {
   "tech_talks": [
     {
-      "date": "2021-10-30",
-      "title": "This is a dummy coming soon event",
-      "presenter": "Karan Singh",
-      "url": "https://www.youtube.com/watch?v=vQF17UBU4RE"
+      "date": "2021-06-02",
+      "title": "Ceph Month 2021: Call for Participation: Ceph Stable Releases Team",
+      "presenter": "Loïc Dachary",
+      "url": "https://youtu.be/KGFfDguicWk"
+    },
+    {
+      "date": "2021-06-02",
+      "title": "Ceph Month 2021: Week 1 Open Discussion",
+      "presenter": "Sage and colleagues",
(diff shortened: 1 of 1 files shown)


src/en/community/community.json
```diff
@@ -1,5 +1,41 @@
 {
   "tech_talks": [
+    {
+      "date": "2021-03-10",
+      "title": "What's New in Pacific Ceph Release",
+      "presenter": "Sage Weil",
+      "url": "https://www.youtube.com/watch?v=PVtn53MbxTc"
+    },
```
liewegas, line 8: should this go at the end so that it's is in reverse chronological order?

zdover23: There's a line in index.html that looks like this:

{% set eventLocaleTag = locale + '-event' %} {% set eventsFuture = collections[eventLocaleTag] | getItemsInFuture | reverse | getItems(4) %}

This line automatically sorts the json objects in community.json into reverse chronological order when the website is rendered.

However (and I don't mean this in a petulant way), if you'd prefer the .json file itself to have these objects sorted by human hands into chronological order, that is a task well within my vim competencies.

Let me know if you want me to hand-sort the json file.

Pete-Robelou: @zdover23 If you want to hand-sort the json file that's fine. It will definitely be easier to edit and maintain if it reads the same way as it renders in the site. I'll leave the decision to you as it's not essential for rendering due to the `reverse` filter.

zdover23: I've sorted them.

Pete-Robelou (approved):
All looks good to me.

src/en/community/community.json
liewegas: :+1:

PR #164, opened 2021-06-11
Card content filtering

### LINKED ISSUE #153: ArticleCard: Content filtering
The `ArticleCard` is used for both `blog-posts` and `press-releases` for all instances wherever they appear in the site.

I spotted a difference between the Blog listing page and the Blog search results page, they display the caption content slightly differently.

Blog listing:
Screenshot 2021-06-09 at 12 05 07

Blog search results:
Screenshot 2021-06-09 at 12 05 53

Ignore the line-break character for now but if you look at the punctuation you'll notice a difference. 

Essentially what is happening here is they're being fed different content. The search results uses `search-output-json` which has been passed through a filter removing punctuation while the listing page uses unfiltered content directly from the `blog-posts` collection.

To bring this inline I think we should update `ArticleCard` itself to filter the content rather than relying on pre-filtered content. That way we know it's output will always be consistent.

Pete-Robelou:
- renamed cleanSearchOutput to cleanCardContent (it's more appropriate)
- added ¶ to list of punctuation
- removed a few characters from the punctuation for CardContent (let's keep an eye on this once more content goes in)
- replaced removeHtml with cleanCardContent in ArticleCard

Closes #153

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -23,7 +23,7 @@ module.exports = function (eleventyConfig) {
   // Filters
   const filtersDir = `./src/_11ty/filters`;
   eleventyConfig.addFilter('chunkByYear', require(`${filtersDir}/chunkByYear.js`));
-  eleventyConfig.addFilter('cleanSearchOutput', require(`${filtersDir}/cleanSearchOutput.js`));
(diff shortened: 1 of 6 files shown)


adamduncan (approved):
👍

PR #169, opened 2021-06-11
ceph.io/users: add telemetry dashboard

zdover23:
This PR adds a link to the Ceph Public Telemetry
Dashboard to the Users page. It also links to the
documentation and cleans up the marketing-speak
that was present previously.

### DIFF
--- a/src/en/users/index.html
+++ b/src/en/users/index.html
@@ -23,23 +23,19 @@ overlaySubMenu: true
     <div class="grid lg:grid--cols-2 grid--gap-14 lg:grid--gap-28">
       <div>
         <p class="standout">
-          Ceph’s community drives innovation and are critical to the successful and performant continuation of Ceph as an industry leading
-          software-defined storage system. Becoming a Ceph developer means you belong to a global community of passionate, talented and
-          inspired developers, with the common goal of making Ceph the best it can be.
+	The Ceph Foundation believes that all storage problems should be solvable with open-source software.
         </p>
-        <a class="button" href="/{{ locale }}/users/documentation/">Get started with Ceph</a>
+        <a class="button" href="https://docs.ceph.com/">Get started with Ceph</a>
       </div>
       <div class="bg-grey-300 p-5 rounded-2">
-        <h3 class="h3">Lorem ipsum</h3>
+        <h3 class="h3">Telemetry</h3>
         <p class="mb-8 lg:mb-12 p">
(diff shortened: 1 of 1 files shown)


src/en/users/index.html
```diff
@@ -23,23 +23,19 @@ <h2 class="h2 mb-12 lg:mb-16">Start using Ceph today</h2>
...
+        <h3 class="h3">Telemetry</h3>
         <p class="mb-8 lg:mb-12 p">
-          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec in dui sed arcu ullamcorper rhoncus. Maecenas lacus ligula, aliquam
-          quis elit vitae, congue viverra libero.
-        </p>
-        <h3 class="h3">Lorem ipsum</h3>
-        <p class="mb-0 p">
-          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec in dui sed arcu ullamcorper rhoncus. Maecenas lacus ligula, aliquam
-          quis elit vitae, congue viverra libero.
+	The telemetry module sends anonymous data about the cluster back to the Ceph developers to understand how Ceph is used and to understand the problems that users might experience.
```
liewegas, line 33: The telemetry module (if enabled by the cluster administrator) ...

src/en/users/index.html
```diff
@@ -23,23 +23,19 @@ <h2 class="h2 mb-12 lg:mb-16">Start using Ceph today</h2>
...
         </p>
+        <p class="mb-8 lg:mb-12 p">
+	Click the button below to see the Ceph Public Telemetry Dashboard. This dashboard reports how many OSDs and active clusters have been registered with Telemetry. You can also see the total capacity of the OSDs being tracked by Telemetry.
```
liewegas, line 36: The telemetry dashboard summarizes information about all Ceph clusters that have opted-in to share anonymized data.  This includes cluster size, Ceph version, which Ceph services are enabled, what types (vendors and models) of storage devices, and more.

zdover23: Text similar to this has been added.

liewegas (approved):

Pete-Robelou (commented):
@zdover23 Just a couple of CSS classes that aren't needed.

`mb-8 lg:mb-12` on the `<p>` tags are only needed when you want to override the default margin bottom of `p`. In this case I don't think additional margins are needed as it's related content just broken down into a couple of paragraphs.

I have another branch where I've been making a few layout tweaks so I'll resolve it there.

PR #170, opened 2021-06-14
Feature/161 node env for dev prod

Pete-Robelou:


adamduncan:
@Pete-Robelou I think the issue re: `process` goes a bit deeper into the JS bundling.

Filters and shortcodes are marked as `dynamicRequireTargets` in our rollup config. This means that all of those JS files (some of which we only intend on running on the Node build) end up in our client bundle. In browser-land, the concept of `process` isn't available, and so throws an error from `bundle-main`.

This comes up because we're importing `site` into various filters (e.g. `formatDate`) in order to grab our "site config" `defaultLocale` value. `formatDate` is then used on the client JS by the blog `ArticleCard`. That then brings with it the `site` module (and its reference to `process`), which yields the error in the browser.

Can we rethink how/what we're bundling here? Feels like we might want to explore:

1. Do we need to `dynamicRequireTargets` all of our filters and shortcodes for the JS bundles? Does that bring with it a lot of modules that are used exclusively at build time?
2. If that (or `ArticleCard` at least) is needed in our client bundles, we might consider moving any build-specific data (e.g. `site.buildTime` and `site.isProduction`) into their own specific data files (`_data/build.js`?) to decouple them from the `site` data we need for our client modules?

Hope that helps

Pete-Robelou:
> * Do we need to `dynamicRequireTargets` all of our filters and shortcodes for the JS bundles? Does that bring with it a lot of modules that are used exclusively at build time?
> 
> * If that (or `ArticleCard` at least) is needed in our client bundles, we might consider moving any build-specific data (e.g. `site.buildTime` and `site.isProduction`) into their own specific data files (`_data/build.js`?) to decouple them from the `site` data we need for our client modules?

I've made our `dynamicRequireTargets` more specific so they only declare the `filters` and `shortcodes` that actually `require`. Based on the [README](https://github.com/rollup/plugins/blob/master/packages/commonjs/README.md#dynamicrequiretargets) is should allow for better optimisation. 

We have a total `23` files between our `filters` and `shortcodes` and now we're just referencing `6` so that should improve things.

I've split up the `site.js` file into 3 separate files now.

**site.js**
```
  title: 'Ceph',
  url: 'https://ceph.io',
  defaultLocale: 'en',
```

**social.js**
```
  facebook: '@cephstorage',
  twitter: '@ceph',
```

**build.js**
```
  buildTime: new Date(),
  isProduction: process.env.NODE_ENV === 'production',
```

So far so good, everything is working fine and we no longer have the `process` warning in the browser.

adamduncan:
Thanks @Pete-Robelou.

> I've split up the site.js file into 3 separate files now.

Perfect 👍 

> I've made our `dynamicRequireTargets` more specific so they only declare the filters and shortcodes that actually require

I think this is close (and technically works), but I'm not sure we're quite targetting things under the hood correctly. We want to target the modules that will be used in the client (i.e. just `shortcodes/ArticleCard.js`) and _its_ required dependencies. So I think we want to be doing that explicitly, and some of the require targets in the PR at 8591177a015a41dc2b66d8edc5f028d4dee09e8b aren't relevant.

I'd consider targetting these specifically, and lifting them as `commonjsOptions` to pass to both the `pluginsES` and `pluginsES5` functions (as they're the same for both bundles):

```js
const commonjsOptions = {
  dynamicRequireTargets: [
    './src/_11ty/filters/cleanCardContent.js',
    './src/_11ty/filters/formatDate.js',
    './src/_11ty/filters/getSingleDigitFromDate.js',
    './src/_11ty/filters/truncate.js',
    './src/_11ty/shortcodes/ArticleCard.js',
  ],
  transformMixedEsModules: true,
};


  // ... 
  commonjs(commonjsOptions),
  //...
```

Ultimately, `dynamicRequireTargets` does include all the modules in our resulting bundles, so moving from the ` './src/_11ty/filters/**/*.js', './src/_11ty/shortcodes/**/*.js'` approach to this very targetted approach does bring our `main` and `search` bundles down from `80kb`/`380kb` to `10kb`/`35kb`, respectively, which is awesome.

adamduncan (changes requested):
Thanks @Pete-Robelou.

> I've split up the site.js file into 3 separate files now.

Perfect 👍 

> I've made our `dynamicRequireTargets` more specific so they only declare the filters and shortcodes that actually require

I think this is close (and technically works), but I'm not sure we're quite targetting things under the hood correctly. We want to target the commonjs modules that will be used in the client (i.e. just `shortcodes/ArticleCard.js`) and _its_ required dependencies. So I think we want to be doing that explicitly, and some of the require targets in the PR at 8591177a015a41dc2b66d8edc5f028d4dee09e8b aren't relevant.

I'd consider targetting these specifically, and lifting them as `commonjsOptions` to pass to both the `pluginsES` and `pluginsES5` functions (as they're the same for both bundles):

```js
const commonjsOptions = {
  dynamicRequireTargets: [
    './src/_11ty/filters/cleanCardContent.js',
    './src/_11ty/filters/formatDate.js',
    './src/_11ty/filters/getSingleDigitFromDate.js',
    './src/_11ty/filters/truncate.js',
    './src/_11ty/shortcodes/ArticleCard.js',
  ],
  transformMixedEsModules: true,
};


  // ... 
  commonjs(commonjsOptions),
  //...
```

Ultimately, `dynamicRequireTargets` does include all the modules in our resulting bundles, so moving from the ` './src/_11ty/filters/**/*.js', './src/_11ty/shortcodes/**/*.js'` approach to this very targetted approach does bring our `main` and `search` bundles down from `80kb`/`380kb` to `10kb`/`35kb`, respectively, which is awesome.

Pete-Robelou:
@adamduncan Ok, I see. I misundertood the `dynamicRequireTargets` a little bit.

I've made the requested updates.

adamduncan (approved):
Nice one :thum

PR #171, opened 2021-06-14
CSS optimisation

### LINKED ISSUE #138: CSS optimisation
- [x] Look at a tool like [css-purge](https://www.npmjs.com/package/css-purge) to automate removal of unused CSS from our production bundle, based on the classes used in the `src` pages and shortcodes, etc.

Alternative:
- https://purgecss.com/

Pete-Robelou:
- npm postcss-purgecss
- configured to look at .html, .njk, .js
- regex for defaultExtractor and safelist

Closes #138

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -4,13 +4,11 @@ module.exports = ({ env }) => ({
     require('postcss-normalize'),
     require('postcss-preset-env')({ stage: 1 }), // https://preset-env.cssdb.org/features#stage-1
     // Other plugins...
-    env === 'production' ? require('cssnano')({ preset: 'default' }) : null, // How to use conditional plugins/options for prod
-    env === 'production'
-      ? require('@fullhuman/postcss-purgecss')({
-          content: ['./src/**/*.html', './src/**/*.njk', './src/**/*.js'],
-          defaultExtractor: content => content.match(/[\w-/:]+(?<!:)/g) || [],
-          safelist: { deep: [/^js-/] },
-        })
-      : null,
+    require('@fullhuman/postcss-purgecss')({
+      content: ['./src/**/*.html', './src/**/*.njk', './src/**/*.js'],
+      defaultExtractor: content => content.match(/[\w-/:]+(?<!:)/g) || [],
+      safelist: { deep: [/^js-/] },
+    }),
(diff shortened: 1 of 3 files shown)


postcss.config.js
```diff
@@ -5,5 +5,12 @@ module.exports = ({ env }) => ({
     require('postcss-preset-env')({ stage: 1 }), // https://preset-env.cssdb.org/features#stage-1
     // Other plugins...
     env === 'production' ? require('cssnano')({ preset: 'default' }) : null, // How to use conditional plugins/options for prod
+    env === 'production'
```
adamduncan, line 8: Thought: Does it make sense to perhaps enable purging in all envs? I.e. I'm thinking should the tooling unintentionally remove some CSS we'd not expected it to, we'd not notice during local development, and it'd ship to production where we might encounter a bug.

I've had a quick test of this locally, and it's very fast when it comes to rebuilding CSS even with the purge. What do you think?

postcss.config.js
```diff
@@ -5,5 +5,12 @@ module.exports = ({ env }) => ({
     require('postcss-preset-env')({ stage: 1 }), // https://preset-env.cssdb.org/features#stage-1
     // Other plugins...
     env === 'production' ? require('cssnano')({ preset: 'default' }) : null, // How to use conditional plugins/options for prod
```
adamduncan, line 7: Thought: I don't _think_ it'll have unitended outcomes, but do you think it's worth running `cssnano` as the last step in the postcss build? A) it is conditional based on env, and B) it's another transformation to the CSS file that could potentially manipulate the raw CSS and interfere with purging?

Seems to be behaving fine either way around. Just a hunch.

adamduncan (approved):
This is great. Massive perf boost here 👍  Wiped out 110kb of unused CSS 😮

postcss.config.js
Pete-Robelou: Yeah, makes sense. Better to spot any CSS issues locally while in dev rather than on the live production site.

postcss.config.js
Pete-Robelou: Like you say it does seem to be working either way but having `cssnano` as the last transformation makes sense.

PR #173, opened 2021-06-14
Husky and lint-staged installation

Pete-Robelou:
- Installed latest versions of `husky`, `lint-staged` and updated `prettier` to latest
- Configured `lint-staged` and `prettier` to look at the same files `js|json|css|md|html`


package.json
```diff
@@ -68,5 +71,8 @@
     "details-polyfill": "^1.1.0",
     "focus-visible": "^5.2.0",
     "lunr": "^2.3.9"
+  },
+  "lint-staged": {
+    "*.--write": "prettier --write"
```
adamduncan, line 76: We want to specify here which files we intend to run `lint-staged` on. Presently this won't lint any files. Interestingly, given this is what the [auto setup](https://github.com/okonet/lint-staged#installation-and-setup) for `lint-staged` gives us, though is it documented anywhere how `*.--write` would work, or that it's something you need to change. I guess the kinda allude to it in their [Configuration](https://github.com/okonet/lint-staged#configuration) docs.

So we want lint-staged to run across the same files Prettier does:

```suggestion
    "**/*.+(js|json|css|md|html)": "prettier --write"
```

(Note: I noticed our prettier is currently including `jsx` files as extensions to lint. We can remove that [from here as well](https://github.com/ceph/ceph.io/blob/develop/package.json#L15))

Also worth noting: Prettier doesn't currently support formatting of `njk` files, so no use in including that in `prettier` script, nor our `lint-staged` config. In future it may. For the time being, we can hope that those primary devs on the project use something like [Nunjucks Template Formatter](https://marketplace.visualstudio.com/items?itemName=okitavera.vscode-nunjucks-formatter) and as we have `"editor.formatOnSave": true`, we should inherit good formatting. Also, `njk` files are primarily template layouts and components, so not as likely to be edited by day-to-day content authors.

adamduncan (changes requested):
Cool, nice start @Pete-Robelou. Good to see how the latest versions of these two packages work together.

Other than the suggested change, I think this is good to go. I've tested on a local branch, and it's successfully reformatted on `pre-commit` across the languages specified.

Do you think it's worth briefly explaining the use of these packages in the main README, should contributors come across it and wonder why a particular file isn't formatted as they expected?

Pete-Robelou:
> Do you think it's worth briefly explaining the use of these packages in the main README, should contributors come across it and wonder why a particular file isn't formatted as they expected?

@adamduncan We could potentially mention that `prettier` doesn't format `njk` files and suggest they use [Nunjucks Template Formatter](https://marketplace.visualstudio.com/items?itemName=okitavera.vscode-nunjucks-formatter). I don't know how useful an explanation of `husky` and `lint-staged` will be.

adamduncan (approved):
Looks good @Pete-Robelou 👍

PR #179, opened 2021-06-16
ansible: Add redirects to nginx config

djgalloway:

### DIFF
--- a/ansible/roles/deploy/templates/redirects.j2
+++ b/ansible/roles/deploy/templates/redirects.j2
@@ -1,0 +1,108 @@
+  # Redirects https://github.com/ceph/ceph.io/issues/129
+  # These get templated into the site.conf in multiple places
+
+  # The /?$ bit at the end of each location will match with or without trailing slash
+  # but nothing beyond that.  e.g.,
+  # ceph.io/get     -> ceph.io/en/users/getting-started
+  # ceph.io/get/    -> ceph.io/en/users/getting-started
+  # ceph.io/get/foo -> 404
+
+  location ~ ^/get/?$ {
+      return 301 /en/users/getting-started;
+  }
+
+  location ~ ^/community/blog/?$ {
+      return 301 /en/news/blog;
+  }
+
+  location ~ ^/careers/?$ {
+      return 301 /en/developers/career-opportunities;
+  }
+
+  location ~ ^/use-cases/cephhome-the-domestication-of-a-wild-cephalopod/?$ {
+      return 301 /en/discover/use-cases;
+  }
+
+  location ~ ^/community/code-of-conduct/?$ {
+      return 301 /en/code-of-conduct;
+  }
+
+  location ~ ^/(trademarks|logos)/?$ {
+      return 301 /en/trademarks;
+  }
+
+  location ~ ^/discover/?$ {
+      return 301 /en/discover;
+  }
+
+  location ~ ^/(use|users)/?$ {
+      return 301 /en/users;
+  }
+
+  location ~ ^/get-involved/?$ {
+      return 301 /en/community;
+  }
+
+  location ~ ^/code/?$ {
+      return 301 /en/developers/code;
+  }
+
+  location ~ ^/(cephalocon|cephdays|events|cephalocon/.*)/?$ {
+      return 301 /en/community/events;
+  }
+
+  location ~ ^/ceph-storage/?$ {
+      return 301 /en/discover/technology;
+  }
+
+  location ~ ^/privacy/?$ {
+      return 301 /en/privacy;
+  }
+
+  location ~ ^/publications/?$ {
+      return 301 /en/news/publications;
+  }
+
+  location ~ ^/community/?$ {
+      return 301 /en/community;
+  }
+
+  location ~ ^/irc/?$ {
+      return 301 /en/community/connect;
+  }
+
+  location ~ ^/jobs/?$ {
+      return 301 /en/developers/career-opportunities;
+  }
+
+  location ~ ^/team/?$ {
+      return 301 /en/community/team;
+  }
+
+  location ~ ^/community/meetings/?$ {
+      return 301 /en/community/meetups;
+  }
+
+  location ~ ^/security/?$ {
+      return 301 /en/security;
+  }
+
+  location ~ ^/(contribute|contributors)/?$ {
+      return 301 /en/developers/contribute;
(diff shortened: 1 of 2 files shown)


adamduncan:
This is great, thank you @djgalloway 👍 

I've run the links through this [301 checker tool](https://www.rankwatch.com/free-tools/url-Redirect-checker), and looks like they're all redirecting nicely ([results](https://docs.google.com/spreadsheets/d/1-8KUFTRIEatncSbXQF-R37gmzAHR-e91SkRLI9hw3Ew/edit?usp=sharing)).

@Thingee Can you please review [the spreadsheet](https://docs.google.com/spreadsheets/d/1HM6g0YDHxaNsBam9wzDPeCrmaW7zR1MKEY2xBuc_1II/edit#gid=900869262) we've put together? David has put in place all the pages we specified. There are a fair few pages on the existing site that don't have an equivalent on the new site (e.g. https://ceph.io/help), or are potentially out of date or no longer maintained (we've not included redirects for these yet). Are there any other important pages we've missed?

The main outstanding pages are the blog posts. They're difficult to target, as they're each housed under their respective tag prefix `/community/:slug` or `/releases/:slug`, `/cds/:slug`, `/ceph-blog/:slug`, `/geen-categorie/:slug`, etc. On the new site, these are better organised by year under `/en/news/blog/:year`. Is there a definitive list of these categories, in which case we can redirect any that start with those paths at least to the new blog homepage?

If there's not a predictable/automated way of capturing these, we may not be able to put redirect rules in place for blog posts. In which case users will end up on a 404 when being linked from old search index results or cross-links, for example. Are we okay with that?

djgalloway:
I just realized these are all using `~` instead of `=`.  This means a request for `ceph.io/privacy/this/path/doesnt/exist` will still return `/en/privacy`.  This also means second-level redirects don't work.  e.g.,  Currently, `beta.ceph.io/community` goes to `/en/community` but `beta.ceph.io/community/meetings`  also goes to `/en/community` which is not what we want.

So we have two options:

1. Change only some of the redirects to `=`.  So in the example of `/privacy/`, where there are no subpaths, we could leave it as `~` and all `/privacy/this/path/doesnt/exist` requests will still go to `/en/privacy`.
2. Change all redirects to `=`.  So `/privacy/` goes to `/en/privacy` but `/privacy/foo/` 404s.

I'm also trying to work out a way to get the blog posts to work.

adamduncan:
@djgalloway Thank you, that's helpful for context. My gut says option 2 there matches my expectations. What do you both think?

djgalloway:
Okay, that's sorted.  I'm working on a way to get blog posts working.

djgalloway:
Got it!  I forked Mike's branch with the blog post import, applies these commits on top of it, then pushed to the `blog-redirects` branch.

e.g.,
https://ceph.io/releases/v15-2-13-octopus-released/
https://blog-redirects.ceph.io/releases/v15-2-13-octopus-released/

dmick:
In private, I expressed bewilderment at the "_" parameter in try_files; to me, that's non-obvious enough that it deserves commentary (I'm assuming it's there as a "never-match" parameter).  Also, nginx's document doesn't make clear that using @<name> as the 'fallback' parameter means that 404 is autogenerated on no match, but that appears to be the case; I'd also comment that, just because it's also nonobvious.

but the changes seem to work.

adamduncan:
@djgalloway Thanks for putting thought into this.

I'm curious about the solution. Should (or is it possible?) `/releases/v15-2-13-octopus-released/` redirect (301) to `/en/news/blog/2021/v15-2-13-octopus-released/`? As to maintain a canonical URL for the posts going forward?

djgalloway:
> @djgalloway Thanks for putting thought into this.
> 
> I'm curious about the solution. Should (or is it possible?) `/releases/v15-2-13-octopus-released/` redirect (301) to `/en/news/blog/2021/v15-2-13-octopus-released/`? As to maintain a canonical URL for the posts going forward?

To be honest, I didn't even notice.  I was just excited to get it working in any way.  I've spent the last hour or so trying to find a way to get `try_files` and a 301 redirect to work together but I don't think it's going to be possible.  They're just different methods.

`try_files` just attempts to return a different file locally.  `301` tells the browser that the requested file(s) is at a different http location is my understanding.

I could `redirect` the blog post string to the search php but that doesn't return the actual post.  https://blog-redirects.ceph.io/en/news/blog/search/?q=v15-2-13-octopus-released

I would like it to redirect but I don't think it's going to be possible without me sinking a bunch of time into scripting something.

adamduncan:
Sure, makes sense. Appreciate you exploring a few options all the same 🙂 

It does help that we have a `canonical` link reference in the new site, so technically crawlers should understand that these legacy URLs aren't the "true" page.

```
<link rel="canonical" href="https://ceph.io/en/news/blog/2021/v15-2-13-octopus-released/">
```

In which case, do we say let's go with the `try_files` approach, as the preferred outcome to having visitors land on 404s?

Question (perhaps for @Thingee): Are there other category directories that we need to be adding for these legacy posts? Looks like we've covered a few of the main I've seen, but I'm not familiar enough with the site to know the full extent.

adamduncan:
@Thingee I've looked through [the posts sitemaps](https://ceph.io/sitemap_index.xml) for the existing site. Based on what's included there, looks as though we have a few more routes the blog posts are currently residing.

@djgalloway As well as those you've already captured, posts also exist under:

- `/ceph-management` ([example](https://ceph.io/ceph-management/introducing-cephadm/))
- `/geen-categorie` ([example](https://ceph.io/geen-categorie/want-to-install-ceph-but-afraid-of-ansible/))
- `/newsletter` ([example](https://ceph.io/newsletter/ceph-community-june-2018/))
- `/rados` ([example](https://ceph.io/rados/new-in-nautilus-pg-merging-and-autotuning/))
- `/rgw` ([example](https://ceph.io/rgw/new-luminous-rgw-metadata-search/))
- `/update` ([example](https://ceph.io/update/new-in-nautilus-device-management-and-failure-prediction/))

I think if we add these to our `try_files` redirects, we'll have everything covered, thank you.

Posts under `/planet/` will not be included in the new site, and so can all redirect to the blog `/en/news/blog/` homepage.

adamduncan:
Thanks for picking up the additions, @djgalloway 👍  Happy for you to merge this one when you're happy

adamduncan (approved):

PR #185, opened 2021-06-17
Update Community pages

Thingee:
Remove filler text and add the Ceph dev mailing list.

### DIFF
--- a/src/_includes/layouts/listing-event-categories.njk
+++ b/src/_includes/layouts/listing-event-categories.njk
@@ -66,19 +66,4 @@
   </ul>
 </section>
 
-<section class="pt-0 section">
-  <div class="bg-grey-300 border-grey-500 border-px border-solid p-5 rounded-2">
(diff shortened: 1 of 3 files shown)


adamduncan (approved):
Looks good, @Thingee 

Side note: The wires/[designs](https://projects.invisionapp.com/share/D8Z2N4X7U6J#/screens/432974007) (p/w: `Jh49k23nf!`) initially proposed some additional channels, IRC, etc. Is that content you'd like to add here?

cc @Pete-Robelou

Pete-Robelou:
I noticed on the [meetups page](src/en/community/meetups/index.html) the second jump link for `meetups` has been updated to an external link, this kinda breaks the UX as they are intended for linking to sections within in the page.

@fishoutoforder, would it be possible to write a small amount of content to maintain the `meetups` page section so we keep the consistent navigation convention for users?

Thingee:
@Pete-Robelou This was on purpuse. The added text is not needed.

PR #186, opened 2021-06-17
Remove ceph-announce mailing list from these pages

Thingee:


Pete-Robelou (approved):
All looks good to me.

We have #190 to create a `shortcode` for the few remaining instances of this section.

PR #193, opened 2021-06-18
doc/cephio: add users->documentation

zdover23:
This PR adds text to the Users->Documentation
page.

### DIFF
--- a/src/en/users/documentation/index.html
+++ b/src/en/users/documentation/index.html
@@ -3,4 +3,31 @@ title: Documentation
 order: 2
 ---
 
-<h1 class="h1">{{ title }}</h1>
+<h1 class="h1">{{ title }}</h1><div>
+
+	<section class="section">
+		  <div class="grid lg:grid--cols-2 grid--gap-14 lg:grid--gap-28">
+
+        <p class="standout">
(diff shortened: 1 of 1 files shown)


src/en/users/documentation/index.html
```diff
@@ -3,4 +3,31 @@
...
+	The upstream Ceph documentation is linked below.
+	</p>
+	<p class="p">
+	The Ceph Install Guide describes how to deploy a Ceph cluster. The
+	cephadm guide describes how to use the cephadm utility to manage your
+	Ceph cluster. If you are consulting the documentation to learn the
+	rules and customs that govern making a pull request against the
+	ceph/ceph Github repository, read the Developer Guide. For more
+	in-depth information about the nature of Ceph, see the Architecture
+	Guide on the page linked below.
```
liewegas, line 25: should `Ceph Install Guide` etc be links?

zdover23: I think if we were on a website in 1998, the answer would be "yes", but in this case we have a button linked to the top-level of the docs and that button appears below this text. I think that adding the links in to this text makes this into a David Foster Wallace novel and takes the whammy out of the button.

If you disagree strongly with this, I'll put the links in. Let me know what you think.

liewegas (approved):

PR #197, opened 2021-06-18
Optimise search index

adamduncan:
- Remove `date` — not sure this field (given it's an ISO date it's not particularly searchable anyway)
- Use spaces instead of empty strings for `shortWords`, `unicode` and `punctuation` replacements. With these as empty strings we were seeing a lot of accidentally combined words in the index, especially in indexed code snippets (e.g. `radosopenpoolconst`). When these are properly separated, the quality of the index actually improves (more keywords), and the size greatly decreases due to compression.
- Removed the `dedupe`ing — this doesn't actually have a great impact on the final index size (due to compression), and may actually decrease the quality of results, as the pages with search terms frequently repeated should rank higher.
- Added `boost` param for `title` and `author` fields, as this also improves relevance of results.

So in total, it takes us from an index of 11MB (2.04MB gzip) to 7.6MB (1.63MB gzip), down ~20% compressed.

Note: Tried a few other tactics for reducing the search index size without impacting the quality of the results. For example, tried omitting long words (e.g. >= 15 characters, which were often a side-effect of splitting/replacing strings in the content) and long number strings (e.g. `1000000`). Because of gzip compression and the repeptitiveness of these strings, it didn't improve the filesize enough to warrant the complexity it introduced to `cleanSearchRaw`.

👀 Also refactored `SearchOutput` slightly, to add error handling and error state, parallelize requests for JSON.

### DIFF
--- a/scripts/search-index.js
+++ b/scripts/search-index.js
@@ -15,10 +15,10 @@ function getData(distDir) {
 }
 
 function buildIndex(articles = []) {
-  let searchIndex = lunr(function () {
+  const searchIndex = lunr(function () {
     this.ref('id');
-    this.field('title');
-    this.field('author');
+    this.field('title', { boost: 10 });
+    this.field('author', { boost: 5 });
     this.field('content');
 
     articles.forEach((article, searchIndex) => {
--- a/scripts/search-index.js
+++ b/scripts/search-index.js
@@ -19,13 +19,12 @@ function buildIndex(articles = []) {
     this.ref('id');
     this.field('title');
     this.field('author');
-    this.field('date');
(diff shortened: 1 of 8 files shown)


Pete-Robelou (approved):
All looks good to me.

Some nice optimisations with the cleanSearchRaw filter, it's interesting how in the pursuit of removing stuff to keep the size down we inadvertantly made it larger and less effective. Doh!

I missed the `boost` functionality of `lunr` so good to see it in there, it nice we can give more important to the blog post titles.

The loading state works well. Maybe a future enhancement to swap it out for a little animation but all good for now.

PR #198, opened 2021-06-18
Updated blog listing pagination order so newest appear first

### LINKED ISSUE #128: Sort order: Blog posts, case studies, articles etc...
The sort order of the blog posts needs to be reversed as we're seeing the oldest posts on the first page.

Screenshot 2021-05-26 at 17 22 15

- [x] Reverse order of blog posts

This may also be applicable for case studies and press releases. It's worth checking this throughout the site.

Pete-Robelou:
A very simple code tweak but worth checking it locally to make sure it's working as expected.

Newest blog posts should be appearing first.

Closes #128

### DIFF
--- a/src/_includes/layouts/listing-blog-posts.njk
+++ b/src/_includes/layouts/listing-blog-posts.njk
@@ -2,6 +2,7 @@
   layout: navigation
   pagination:
     size: 12
+    reverse: true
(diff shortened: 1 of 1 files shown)


adamduncan (approved):
Thanks 👍

PR #200, opened 2021-06-18
Removed Press Releases and Planet Ceph sections

### LINKED ISSUE #181: Remove Press release and Planet Ceph sections for launch
- [x] "Releases" posts will be imported into Blog (#174), as per existing site. Not sure we have need for Releases at time of go live
    - For context, was spec'd as: _"Ceph foundation announcements (internal), Major announcements from members"_. Will leave to group to decide if/when it's revived
- [x] Planet Ceph section can go entirely, as also won't have content for launch

Pete-Robelou:
I also updated the `News` page to remove any links to `Press Releases` or `Planet Ceph`.

Closes #181


adamduncan (approved):
👍

PR #202, opened 2021-06-20
Removing training services until it exists

Thingee:


Pete-Robelou:
@Thingee 

I noticed this is the same as #203 but also removes the `Dive into Ceph` section with the promo modules. I couldn't see a reason to remove the promo modules so I've closed this PR in favour of #202 .

PR #203, opened 2021-06-20
Removing career and training services until we have content.

Thingee:


Pete-Robelou (approved):
@Thingee All looks good.

I did notice this has the same changes as #202 but keeps the `Dive into Ceph` section with the promo modules. I couldn't see a reason to remove the promo modules so I'll merge this PR and close off #202.

Issue #208, opened 2021-06-21
Developers landing page requires additional content/editing


fishoutoforder:
The Developers page has the following remaining content gaps:

1. Under **Connect with the community** section, there is one Lorem ipsum box.
2. Under **Connect with the community** section, there is a box describing/linking to career services, will need to replace with something else (as outlined in #203, we're not launching with the careers pages).
3. Under the **Resources for developers section**, there are three boxes with lorem ipsum text (the wireframe suggests "Career opportunities", "Tech talks" and "Upcoming events" here.

PR #211, opened 2021-06-22
Doc/208 - Replacement of all lorem ipsum on developers page

fishoutoforder:
Addresses #208 by replacing all lorem ipsum text with resources for developers and blurbs on community features.


adamduncan (approved):
Thanks @fishoutoforder 👍

PR #215, opened 2021-06-22
Cross-browser checks

adamduncan:
Only update that arose based on browser testing (Edge, makes use of `manifest.json`, which wasn't present in `dist` files).

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -81,6 +81,7 @@ module.exports = function (eleventyConfig) {
   // Transforms
 
   // Passthrough copy
+  eleventyConfig.addPassthroughCopy('./src/assets/**/*.json');
 
   // Plugins
   eleventyConfig.addPlugin(eleventyNavigationPlugin);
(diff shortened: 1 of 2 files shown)


adamduncan:
I've run the site through BrowserStack to test against real devices, both mobile and desktop, based on latest 2 versions. Things are all looking good.

Other than the known layout degradations of <= IE11, which [we're not supporting](https://docs.google.com/document/d/1Pp5qt1D4_IIIPj1juO_iTbm7VZ1DBQySj_U4uHVOy78/edit?pli=1#heading=h.nlu2rs6indef), things are looking good.

Not a lot can technically go wrong with the site, from a functional perspective. The JS features (blog search, nav and menu interactions), so we should feel confident the JS bundling and transpiling is working well.

adamduncan:
@Pete-Robelou, did you want to have a final sweep too, should we need to add anything to this PR?

Pete-Robelou:
> @Pete-Robelou, did you want to have a final sweep too, should we need to add anything to this PR?

I've had a look through FireFox, Chrome and Safari and didn't find any issues.

Pete-Robelou (approved):

PR #218, opened 2021-06-22
Use social usernames as is, adding @ where needed

adamduncan:
Social share for Twitter have a subtle issue.

OG tag for `twitter:site` is expecting a twitter `@username` ([docs](https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started)), while the social share intent link (from Blog posts), is expecting just the username itself, no `@`. Including it leads to a double `@@`, and username is not linked appropriately:

Screenshot 2021-06-22 at 12 00 59

This PR takes `@` usage out of `_data/social`, so we can explictly include it in templates only where needed.

`social.facebook` isn't being used yet, as far as I can see.

### DIFF
--- a/src/_data/social.js
+++ b/src/_data/social.js
@@ -1,4 +1,4 @@
 module.exports = {
-  facebook: '@cephstorage',
-  twitter: '@ceph',
+  facebook: 'cephstorage',
+  twitter: 'ceph',
 };
--- a/src/_data/social.js
+++ b/src/_data/social.js
@@ -1,4 +1,5 @@
 module.exports = {
   facebook: 'cephstorage',
   twitter: 'ceph',
+  youtube: 'UCno-Fry25FJ7B4RycCxOtfw',
 };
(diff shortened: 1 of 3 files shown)


Pete-Robelou (approved):
Nice little update.

I know we've talked about this but do you think it's also worth updating the two references of `ceph` and `cephstorage` in `site-footer.njk` to use the values from `_data/social.js` ?

adamduncan:
>  worth updating the two references of ceph and cephstorage in site-footer.njk to use the values from _data/social.js ?

I was going to leave this, as a lower priority than other tasks. But while we're here, we might as well nudge along the improvement a little.

I've chosen not to loop over all social properties dynamically, as it'd involve some more complexity re: icons and link text labelling, that we don't get from our simple `_data/social` object map at the moment. Updated in 0090572

Pete-Robelou:
I'd say let's get it merged 💪

PR #224, opened 2021-06-22
HTML minify

### LINKED ISSUE #220: Minify HTML
Minify all HTML output files as part of 11ty build config.

E.g. from [this reference](https://mattclaffey.medium.com/how-to-reduce-you-html-output-in-11ty-ec0c358be2ac), or from one of our other 11ty sites:

```js
const htmlmin = require('html-minifier');
  
 //...

    eleventyConfig.addTransform('htmlmin', (content, outputPath) => {
      if (outputPath.endsWith('.html')) {
        let minified = htmlmin.minify(content, {
          collapseWhitespace: true,
          minifyCSS: true,
          minifyJS: true,
          removeComments: true
        });
        return minified;
      }
      return content;
    });
```

Suggest we perhaps do this for all environments, as opposed to just for `production`, just to get a better steer on any potential pitfalls or minification quirks during development.

This will boost TTFB of responses for the inital load of HTML pages. For example, the homepage should go from 17kb to 14kb. Every lil' helps on that first request.

Pete-Robelou:
- Installed html-minifier package
- Created transforms folder
- Fixed a few syntax issues on /community/team

Closes #220

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -2,11 +2,11 @@ const fs = require('fs');
 
 // Plugins
 const eleventyNavigationPlugin = require('@11ty/eleventy-navigation');
-const sitemap = require('@quasibit/eleventy-plugin-sitemap');
+const build = require('./src/_data/build');
 const i18n = require('eleventy-plugin-i18n');
 const markdownIt = require('markdown-it');
 const markdownItAnchor = require('markdown-it-anchor');
-const build = require('./src/_data/build');
+const sitemap = require('@quasibit/eleventy-plugin-sitemap');
 const translations = require('./src/_data/i18n');
 
 module.exports = function (eleventyConfig) {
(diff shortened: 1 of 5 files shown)


.eleventy.js
```diff
@@ -2,11 +2,11 @@ const fs = require('fs');
 
 // Plugins
 const eleventyNavigationPlugin = require('@11ty/eleventy-navigation');
-const sitemap = require('@quasibit/eleventy-plugin-sitemap');
+const build = require('./src/_data/build');
```
adamduncan, line 5: Can we omit this if we're not doing env-specific minification?

adamduncan (approved):
End result here (https://220-minify-html.ceph.io/en/) looks good to me 👍 

If you've not noticed any undesired side-effect, happy to merge

.eleventy.js
Pete-Robelou: It's being used for the sitemap plugin.

```
eleventyConfig.addPlugin(sitemap, {
    sitemap: {
      hostname: build.isProduction ? 'https://ceph.io' : 'https://develop.ceph.io',
    },
  });
```

adamduncan: Ah, my bad. Was just confused by the shift in these diffs. Makes sense 👍  Thanks for clarifying

Pete-Robelou: I should probably have mentioned that I sorted the plugins into alphabetical order 🤪

PR #225, opened 2021-06-22
UI fixes

### LINKED ISSUE #190: Shortcode: Subscribe to Ceph Announcements

Pete-Robelou:
- Added CSS utility w-full to images on /community/team/
- Added link to Subscribe to Ceph Announcements section

Closes #190

### DIFF
--- a/src/en/community/team/index.html
+++ b/src/en/community/team/index.html
@@ -123,7 +123,7 @@ order: 5
 
       <div class="pt-0 section">
         <div class="grid md:grid--cols-56-auto lg:grid--cols-72-auto">
-          <img alt="Sage Weil" class="rounded-2 to-md:max-w-56" loading="lazy" src="/assets/bitmaps/photo-sage-weil.jpg" />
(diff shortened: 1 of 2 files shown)


adamduncan (approved):
All good here, @Pete-Robelou.

I'm pondering the `w-full` approach. I know some earlier work with responsive images (to ensure responsive images didn't automatically stretch to fill their container), and down with the opt-into-full-width approach when images are in content, but wondering about the need for "templated" images to all have to explicitly/manually add this `w-full` class too.

Perhaps it's just the way things are with the atomic CSS approach (i.e. it'd usually be catered to with predefined "object" styles, e.g. the good ol' media object)?

Pete-Robelou:
It came about because of the fix we implemented that removed the blanket declaration of `width: 100%` on `img`.

It's now an opt in scenario but these images need to be full width of their parent and it's part and parcel of the atomic CSS approach. 

There's noting stopping us wrapping up more styles in an object/component approach in the future.

PR #230, opened 2021-06-23
Changes to ansible post go-live

djgalloway:


adamduncan:
Niceee. Thanks @djgalloway 

Will this ensure that the production build for `ceph.io` is running `npm run build:production` for the `main` deployment?

Or will we need to amend the `npm run build` references [here](https://github.com/ceph/ceph.io/blob/main/ansible/roles/deploy/tasks/main.yml#L74) and [here](https://github.com/ceph/ceph.io/blob/main/ansible/roles/deploy/tasks/main.yml#L164)?

(Just asking as we can currently see a `development` environment setting for `nofollow, noindex` on the production site)

Thanks

djgalloway:
> Or will we need to amend the `npm run build` references [here](https://github.com/ceph/ceph.io/blob/main/ansible/roles/deploy/tasks/main.yml#L74) and [here](https://github.com/ceph/ceph.io/blob/main/ansible/roles/deploy/tasks/main.yml#L164)?

Good catch.  Added commits to ensure the proper build env.  Ideally, this playbook will never have to get run again  :innocent: but it's there just in case.

djgalloway:
Just discovered why it was in development mode.  Should be fixed shortly.

adamduncan:
@djgalloway Notice the development feature (robots `noindex` meta tag) we were looking out for has gone from production 👍 

Does that mean this fix is in place and working as intended?

Happy for you to merge this as needed.  Thanks

adamduncan (approved):

PR #232, opened 2021-06-23
Updating Ceph mailing list links to correct location

Thingee:


idryomov:
Just noticed this too.

idryomov (commented):
With this change ceph-announce subscribe and archive links still point to ceph-users.

robbat2 (approved):

PR #257, opened 2021-08-12
src/_includes/layouts/_base: add motomo tracking

liewegas:
Signed-off-by: Sage Weil <sage@newdream.net>

### DIFF
--- a/ansible/roles/deploy/defaults/main.yml
+++ b/ansible/roles/deploy/defaults/main.yml
@@ -17,4 +17,4 @@ packages:
   - net-tools
 node_version: 14.2.0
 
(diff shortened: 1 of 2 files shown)


adamduncan:
Nice one, folks 👍 And an async snippet. Love it! 🙂 

I'd suggest we might `dns-prefetch` for the `ceph.matomo.cloud` ([reference](https://matomo.org/blog/2017/04/important-performance-optimizations-load-piwik-javascript-tracker-faster/)) domain to improve performance. Will make a PR.

adamduncan:
Re: performance tests: https://github.com/ceph/ceph.io/pull/259#issuecomment-900979085

PR #269, opened 2021-08-30
Add logo page

Thingee:


liewegas (approved):
The images seem to have width/height values that mess with the original images' aspect ratio.  Is that fixable?

Thingee:
> The images seem to have width/height values that mess with the original images' aspect ratio. Is that fixable?

Let me know if this is better.

liewegas:
> > The images seem to have width/height values that mess with the original images' aspect ratio. Is that fixable?
> 
> Let me know if this is better.

luminous, nautilus, kraken, jewel, hammer, firefly, dumpling still look a bit off
also the first row of ceph logos

it's annoying that this requires manually scaling the dimensions :(

Pete-Robelou:
> > > The images seem to have width/height values that mess with the original images' aspect ratio. Is that fixable?
> > 
> > 
> > Let me know if this is better.
> 
> luminous, nautilus, kraken, jewel, hammer, firefly, dumpling still look a bit off
> also the first row of ceph logos
> 
> it's annoying that this requires manually scaling the dimensions :(

@liewegas If you remove the `width` and `height` attributes from each of the images they'll display at the correct aspect ratio. There are some CSS rules so when an image doesn't have a `width` and/or `height` they use `width: auto` and `height: auto`, this means they are responsive but will never display larger than their native size while maintaining the correct aspect ratio. 

The intention was to give the editor the ability so override the native size by using `width` and/or `height` attributes.

I also noticed a closing `</p>` on the first paragraph but no opening `<p>`. As it's a markdown file you don't really need any HTML tags except for the grids to display the logos.

Thingee:
> > > > The images seem to have width/height values that mess with the original images' aspect ratio. Is that fixable?
> > > 
> > > 
> > > Let me know if this is better.
> > 
> > 
> > luminous, nautilus, kraken, jewel, hammer, firefly, dumpling still look a bit off
> > also the first row of ceph logos
> > it's annoying that this requires manually scaling the dimensions :(
> 
> @liewegas If you remove the `width` and `height` attributes from each of the images they'll display at the correct aspect ratio. There are some CSS rules so when an image doesn't have a `width` and/or `height` they use `width: auto` and `height: auto`, this means they are responsive but will never display larger than their native size while maintaining the correct aspect ratio.
> 
> The intention was to give the editor the ability so override the native size by using `width` and/or `height` attributes.
> 
> I also noticed a closing `</p>` on the first paragraph but no opening `<p>`. As it's a markdown file you don't really need any HTML tags except for the grids to display the logos.

Thanks Pete!

PR #299, opened 2021-11-05
community/events: adding Cephalocon 2022

Thingee:
Also, hide the registration button if there is no registration link yet.


Thingee:
@adamduncan, the cropping on the community/events page isn't quite right here. Not sure what can be done there?

Pete-Robelou:
> @adamduncan, the cropping on the community/events page isn't quite right here. Not sure what can be done there?

@Thingee I had a look and it's because the image isn't 16x9 aspect ratio.

I had a little play in Photoshop to make it 16x9, hopefully this helps.

cephalocon-2022-hero

adamduncan:
Thanks @Pete-Robelou 👍

PR #301, opened 2021-11-05
assets: update cephalocon 2022 image aspect ratio

liewegas:
Signed-off-by: Sage Weil <sage@newdream.net>


liewegas:
better! https://fix-cephalocon-image.ceph.io/en/community/events/

PR #377, opened 2022-03-30
community/tech-talk: Add Running Teuthology Tech Talk

Thingee:


src/en/community/community.json
```diff
@@ -4,7 +4,7 @@
       "date": "2022-03-24",
       "title": "Ceph Tech Talk: How Teuthology Works",
```
neha-ojha, line 5: Could we add "locally" at the end?

PR #395, opened 2022-05-17
Adding tags to all release blog posts

drakonstein:
I like to use the tags when browsing the Ceph blog, but I noticed that the "release" only has 3 releases that have been tagged (https://ceph.io/en/news/blog/category/release/). I figured it'd be more effective to put together this PR to kick things off. I'm not sure how these are generated right now, but I assume there's a script or a template in play. Modifying that template to include these tags could keep this up to date from now on.

I updated all releases to have the "release" tag and all releases associated with a named version of Ceph with that name.


Thingee:
LGTM thanks @drakonstein!

PR #421, opened 2022-07-26
Engineering Manager, Ceph & Distributed Storage

javacruft:
Add new engineering manager role at Canonical.


Thingee (approved):
Thanks for adding!

PR #494, opened 2023-02-06
Ceph Day NYC 2023 correct AD employer

anthonyeleven:
The current schedule lists Intel as my employer.  This is not accurate.

When we did the work, I worked for Intel
When it was merged, I worked for Solidigm (Intel spinoff) and the deck uses their template.
I now work for Index Exchange, who is not involved in this work.

Thus, I think it's best to not list an employer for me, to reduce confusion and so that Index
doesn't take issue with it.

Signed-off-by: Anthony D'Atri <anthonyeleven@users.noreply.github.com>


dvanders:
@anthonyeleven could you please re-push this? There was a bug elsewhere in the site preventing this from compiling properly.

anthonyeleven:
Ack, pushed.

PR #539, opened 2023-03-20
community/jobs: Updating jobs page for march 2023

Thingee:


anthonyeleven (approved):
Suggest adding an explicit column for location

PR #627, opened 2023-11-15
Update jobs.json with GoDaddy opening

bstillwell-godaddy:


dvanders:
@anthonyeleven @zdover23 any idea why this didn't render a test page?

dvanders (approved):

zdover23:
> @anthonyeleven @zdover23 any idea why this didn't render a test page?

I don't know if this will come across as an insolent answer, but I think it didn't render a test page because this is ceph.io and not docs.ceph.com, and ceph.io doesn't run renderthedocs builds.

zdover23:
...on the other hand, it didn't run the "Compiling Site" check. I'm not sure why.

PR #683, opened 2024-05-14
Update index.html

RohitYandigeri:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=b_RE9kh)


ljflores:
jenkins retest

ljflores:
Hey @RohitYandigeri this change looks good, except we have an issue right now where the site compile check is not run for people outside the ceph org. I raised the change again in #760, which is your same change, but this is so we can verify that it compiles correctly.

PR #694, opened 2024-06-07
community/ambassadors: Update Danny's bio and picture

dabukalam:


neha-ojha:
Looks like the site didn't compile on this PR, will test a revert

neha-ojha:
The revert https://github.com/ceph/ceph.io/pull/695 has started compiling, @dabukalam I think there is something about the format that is making the check fail silently.

PR #723, opened 2024-08-09
src/en/community/events/2024/ceph-developer-summit-tentacle: move RBD session again

idryomov:
... from Monday to Tuesday, same time.  Note that this extends the CDS by one day.


neha-ojha (approved):
https://wip-move-cds-rbd-session-again.ceph.io/en/community/events/2024/ceph-developer-summit-tentacle/

PR #741, opened 2024-11-02
Adjust Foundation Member Logos

jasonvds:
All logos were set to 200x200 during initial build of ceph.io per 8b6cd70d595a0ec7007e2b7bdaf83fbf7adebe97

Subsequent logo uploads have been across a variety of sizes which affects the way they are rendered and pages, such as https://ceph.io/en/foundation/members/ are affected.

Logos have been resized, to a minimum of 72dpi and set to 200x200px


ljflores:
Hey @jasonvds, this change looks good. However, the site compile check doesn't run for contributors outside the ceph org right now -- while we're fixing that, I raised #741, which is the same as your PR, but so we can confirm it compiles.

gauravsitlani:
jenkins retest this please

ljflores:
jenkins retest

ljflores:
Hi @jasonvds, the site compile check won't run for people outside the ceph org, so I raised the PR again here: #762.

PR #771, opened 2025-01-02
Update some dependencies with npm audit fix

zmc:


dmick:
npm is pretty old on the build host, which I suspect is the cause of this failure.  Trying the same stopgap fix to package-lock.json with npm on the build host.

PR #833, opened 2025-03-13
Update node dependencies

Thingee:
* Update main eleventy package.
* Update some severe outdated dependencies.
* Avoid templateContent after page generation.
* Image related plugins require a plugin name to be specified.
* Some image manipulation related packages need a default function specified.

Fixes: #823, #775, #776, #778, #779, #781, #782


jmundack:
@nizamial09 / @dmick  - mind taking a quick look when you have a min?

Issue #748, opened 2024-12-06
Unable To Run Site Locally


bensternthal:
On a fresh install, running the site locally via ```npm run dev``` results in the following error when I view a page in a browser:

```
[Browsersync] Serving files from: dist
Error [ERR_HTTP_HEADERS_SENT]: Cannot write headers after they are sent to the client
    at ServerResponse.writeHead (node:_http_server:345:11)
    at res.writeHead (/Users/ben/Documents/Dev/ceph.io/node_modules/resp-modifier/index.js:140:27)
    at /Users/ben/Documents/Dev/ceph.io/.eleventy.js:139:15
    at call (/Users/ben/Documents/Dev/ceph.io/node_modules/connect/index.js:239:7)
    at next (/Users/ben/Documents/Dev/ceph.io/node_modules/connect/index.js:183:5)
    at SendStream.error (/Users/ben/Documents/Dev/ceph.io/node_modules/serve-static/index.js:121:7)
    at SendStream.emit (node:events:519:28)
    at SendStream.error (/Users/ben/Documents/Dev/ceph.io/node_modules/send/index.js:270:17)
    at SendStream.onStatError (/Users/ben/Documents/Dev/ceph.io/node_modules/send/index.js:421:12)
    at next (/Users/ben/Documents/Dev/ceph.io/node_modules/send/index.js:764:28)
Error [ERR_HTTP_HEADERS_SENT]: Cannot write headers after they are sent to the client
    at ServerResponse.writeHead (node:_http_server:345:11)
    at /Users/ben/Documents/Dev/ceph.io/.eleventy.js:139:15
    at call (/Users/ben/Documents/Dev/ceph.io/node_modules/connect/index.js:239:7)
    at next (/Users/ben/Documents/Dev/ceph.io/node_modules/connect/index.js:183:5)
    at SendStream.error (/Users/ben/Documents/Dev/ceph.io/node_modules/serve-static/index.js:121:7)
    at SendStream.emit (node:events:519:28)
    at SendStream.error (/Users/ben/Documents/Dev/ceph.io/node_modules/send/index.js:270:17)
    at SendStream.onStatError (/Users/ben/Documents/Dev/ceph.io/node_modules/send/index.js:421:12)
    at onstat (/Users/ben/Documents/Dev/ceph.io/node_modules/send/index.js:727:26)
    at FSReqCallback.oncomplete (node:fs:197:21)
[...]

PR #847, opened 2025-03-25
Fix Eleventy Configuration errors

ONKARBH:
"As requested, moving Eleventy configuration fixes to a separate PR."

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -1,8 +1,9 @@
 const fs = require('fs');
+const path = require('path');
 
 // Plugins
 const eleventyNavigationPlugin = require('@11ty/eleventy-navigation');
-const build = require('./src/_data/build');
+const build = require('./src/_data/build') || {};
 const i18n = require('eleventy-plugin-i18n');
 const markdownIt = require('markdown-it');
 const markdownItAnchor = require('markdown-it-anchor');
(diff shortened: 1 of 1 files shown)


.eleventy.js
```diff
@@ -1,87 +1,112 @@
...
+    }
+    return fs.existsSync(filePath) ? require(filePath) : () => {};
+  };
+
   // Collections
-  const collectionsDir = `./src/_11ty/collections`;
-  eleventyConfig.addCollection('primary', require(`${collectionsDir}/primary.js`));
-  eleventyConfig.addCollection('sitemap', require(`${collectionsDir}/sitemap.js`));
+  const collectionsDir = './src/_11ty/collections';
+  eleventyConfig.addCollection('primary', requireIfExists(`${collectionsDir}/primary.js`));
```
Thingee, line 30: Why is requireIfExists necessary if it's in the main code base?

Thingee:
I was able to verify that around line 161 helps with issue #748.

@ONKARBH any other errors you can note in the PR would be helpful.

ONKARBH:
> More information on why these are errors in the commit message would be helpful for future reference.

- Used `requireIfExists` to prevent errors if the specified collection file is missing.  
- This ensures that the site does not break during the build process if a collection is not available.  
- Verified that these changes help resolve issue #748, allowing the site to run locally without errors.

PR #869, opened 2025-04-15
Add NB to tiering blog crossposts

anthonyeleven:
For whatever reason this functionality has yet to appear in Squid.  Add a note to this effect that makes sense both today and in the future.


zdover23:
LGTM. I'm going to merge this and then make another commit to add more durable temporal information.

zdover23 (approved):

PR #874, opened 2025-04-22
blog: fix old pre formatting markdown

dvanders:
This tests changing the markdown format, so that old blog articles are rendered correctly.


anthonyeleven:
@dvanders bump?

PR #941, opened 2025-09-24
Join me at Cephalocon blog post part 1

anthonymicmidd:


anthonyeleven:
>Fixed the formatting of the jpg to make the people look human :)

Now now, we shouldn't deceive the public.  We must wave our tentacles with pride.

anthonyeleven:
So, I adjusted the tone curve of the image in this PR, and since it is a highly compressed phone photo I had to apply a bit of noise reduction to mitigate the low dynamic range of the dark areas.  My git-fu is not strong enough to update  the file in the commit, but here's what I came up with, you might download it and amend the PR accordingly.


alll-ambassadors-edited

anthonyeleven (approved):

PR #970, opened 2025-11-20
Blog: Fast Erasure Coding for Tentacle Performance Updates

lee-j-sanders:
This blog discusses the performance updates of the new Fast EC (Erasure Code) feature that is in Tentacle.

I have built and viewed the blog updates inside a built container.


jakobdettner:
Nice article. Thank you!
In: "Choosing an Erasure Code Profile" for the storage overhead it should be '(K+M) / K' instead of '(K+M / M)'.

PR #988, opened 2025-12-16
blog: Ending Support for some Erasure Code Plugins

jamiepryde:


jamiepryde:
Hi @anthonymicmidd and @lee-j-sanders, please could I get approval to publish this blog post. I sent an email to the dev and user mail lists on 16 December asking for comments / concerns and have not received any opposition to the proposal we've laid out, so I think we are ready to publish it now.

PR #1065, opened 2026-04-22
Ceph days and talks

anthonymicmidd:


src/assets/bitmaps/image (3).png
jmundack, line 1: why do we have 2 copies of the image?

anthonymicmidd: I renamed this image, but I didn't remove it. I'll do that now

markhpc (approved):
LGTM assuming all of the dates are correct and such.

anthonyeleven (approved):

PR #1085, opened 2026-05-12
update the team list - add nvmeof component leader

caroav:


src/en/community/team/index.html
```diff
@@ -61,6 +61,10 @@ <h3 class="h3">Component Leads</h3>
               <td><a class="a" href="#yingxin-cheng">Yingxin Cheng</a></td>
               <td>Seastore</td>
             </tr>
+            <tr>
+              <td><a class="a" href="#Aviv Caro">Aviv Caro</a></td>
+              <td>Nvme-oF</td>
```
idryomov: Nit: Canonical capitalization?
```suggestion
              <td>NVMe-oF</td>
```

idryomov (commented):
I think this is supposed to be an alphabetical list, so it should be placed between Dashboard and RADOS.

src/en/community/team/index.html
```diff
@@ -61,6 +61,10 @@ <h3 class="h3">Component Leads</h3>
               <td><a class="a" href="#yingxin-cheng">Yingxin Cheng</a></td>
               <td>Seastore</td>
             </tr>
+            <tr>
+              <td><a class="a" href="#Aviv Caro">Aviv Caro</a></td>
```
idryomov: The link is going to be dead anyway, so I'd make it explicit:
```suggestion
              <td><a class="a" href="#">Aviv Caro</a></td>
```

caroav: Looking at the other components it looks like :
`<td><a` class="a" `href="#Name">Name</a></td>`
What am I missing?

idryomov: Most of them are dead links.  This doesn't matter much, but there are also many examples of `href="#"`.

jmundack:
FYI - https://tracker.ceph.com/issues/76442

there is some work being done to move the component leads list into docs.ceph.com (along with the code/technical things)

so can we close this for now - and open a PR to that list?

anthonymicmidd (approved):
LGTM

PR #1086, opened 2026-05-12
Show upcoming events in the correct order

jmundack:
Currently future events show the next event at the end. This fixes the order so that upcoming events are in the expected order.

- removed unnecessary calls to `reverse`
- keep list of future events sorted at source

image

upcoming events shows the "next event" first. Past events shows the "most recent one" first


anthonymicmidd (approved):
WOW! Thanks for this fix!

PR #1111, opened 2026-06-14
fix: Fix reflected XSS in blog search and emit search

RaminNietzsche:
## Summary

Fixes reflected XSS in blog search and replaces hand-built JSON templates
that could break the search index when post titles contain quotes or special
characters.

### 1. Reflected XSS via `?q=` parameter

The raw search query from `urlParams.get('q')` was interpolated into HTML
and assigned via `innerHTML` without escaping. A crafted URL could execute
arbitrary JavaScript in the visitor's browser.

### 2. Fragile search JSON generation

`search-raw.html` and `search-output.html` built JSON by string concatenation.
Post titles or content with `"` or control characters produced invalid JSON,
causing `scripts/search-index.js` to fail at build time.

### 3. Broken fetch mode (same as #1091)

`mode: 'no-cors'` on same-origin fetch calls returned opaque responses,
making `.json()` fail on every search query.

## Changes

- `src/js/search-output.js`: add `escapeHtml()`, use `safeQuery` in output;
  remove `no-cors` / `credentials: 'include'` from fetch
- Replace `search-raw.html` / `search-output.html` with `.11ty.js` templates
  that emit JSON via `JSON.stringify()`
- Use locale-aware collection tag (`{locale}-blog-post`)

## Overlap with upstream

This overlaps with open PR #1091 (XSS + fetch fix in `search-output.js`).
This PR adds safe JSON generation, which #1091 does not cover.
screenshot


RaminNietzsche:
@ceph/ceph-website

RaminNietzsche:
Gentle ping @ceph/ceph-website — still looking for a review when you have a moment.

This fixes reflected XSS in blog search (`?q=` → HTML), emits search JSON via `JSON.stringify`, and drops broken `no-cors` fetch. Overlaps with open #1091 on the XSS/fetch side; this PR also covers safe JSON generation for the search index.

Also need `ok to test` so CI can run. Thanks!

RaminNietzsche:
@anthonymicmidd @anthonyeleven
Is it possible to check this PR if you have time?

PR #1123, opened 2026-06-24
Rename Ceph Days London presentation PDFs

anthonymicmidd:
Updated the file names for each PDF. Removing spaces and special characters. Also updated the path in the index file


anthonyeleven (approved):
Strong approval.  Filenames with embedded spaces are just cruel.

PR #1159, opened 2026-07-27
Update Slack invite link in index.html

anthonymicmidd:


jmundack (approved):

jmundack:
with our new slack pro - @anthonymicmidd  - would this be the last time we have to do this?
(ie - is the new link a "non-expiring" link?)

anthonymicmidd:
> with our new slack pro - @anthonymicmidd - would this be the last time we have to do this? (ie - is the new link a "non-expiring" link?)

@jmundack  Yes, this link is set to non-expiring. But so was the last link. I'm not sure how/why the previous one stopped working. I'll ask Ben to verify this.

PR #1163, opened 2026-07-28
Add release signal to Home page and reframe AI/ML messaging

eameh-LF:


src/en/index.html
```diff
@@ -14,11 +14,20 @@ <h1 class="h1 visually-hidden">Ceph.io {{ title }}</h1>
...
-      <a class="button" href="/{{ locale }}/users/getting-started/">Get Started with Ceph</a>
+      <a class="button" href="/{{ locale }}/discover/">Discover Ceph</a>
     </div>
   </div>
 </section>
 
+<div class="bg-navy py-4 section--full">
+  <div class="wrapper text-center">
+    <p class="color-white mb-0 p">
+      <strong>Latest release: Ceph v20.2.1 Tentacle</strong> &mdash; April 2026 &nbsp;
```
anthonyeleven: Why 20.2.1 here?  20.2.3 is out in a couple of days.

eameh-LF: Good catch — v20.2.3 isn't in the blog yet (no post under src/en/news/blog/2026/). Once the release post is published, I'll update the callout to point to v20.2.3.

anthonyeleven:
Since it is incipient, I suggest holding off on this merge until it's released.  I've added you to the release channels so you should be able to seen when releases happen.

src/en/index.html
eameh-LF: Done: v20.2.3 shipped today and its release post is now on main, so the callout points at v20.2.3 (August 2026). Branch is rebased onto current main so the link resolves in the preview.

eameh-LF:
v20.2.3 is released and the blog post is live, so the hold-until-released concern should be resolved. The callout now points at the v20.2.3 post and the branch is rebased onto current main. Ready for another look.

PR #1172, opened 2026-08-02
Fix Slack card markup on Connect page

eameh-LF:
## Summary

Scope reduced per Foundation guidance: IRC is community-run rather than Foundation-supported, so ceph.io does not promote it (the docs Get Involved page continues to list the channels). This PR now only fixes the Slack card markup on the Connect page:

- `<p lass="p mb-0">` to `<p class="p mb-0">` (the paragraph styling classes were silently not applied)
- Heading level corrected from `h3 class="h2"` to `h3 class="h3"` with consistent spacing
- Subscribe button wrapped to match the page's other cards

No IRC content is added.

## Test plan

- [x] Slack card renders with proper paragraph styling and heading scale
- [x] No IRC mention anywhere on the page

### DIFF
--- a/src/en/community/connect/index.html
+++ b/src/en/community/connect/index.html
@@ -8,9 +8,11 @@ order: 4
     <div class="grid lg:grid--cols-2 grid--gap-14 lg:grid--gap-28">
     <div>
     <div class="bg-grey-300 mb-8 lg:mb-12 p-5 rounded-2">
-      <h3 class="h2">Ceph Foundation Slack</h3>
-      <a class="button" href="https://join.slack.com/t/ceph-storage/shared_invite/zt-453jwr8e9-wu~QvC0jcH4oZBug3oMgdg">Subscribe</a>
-     <p lass="p mb-0">
+      <h3 class="h3 mb-4">Ceph Foundation Slack</h3>
+      <div class="mb-4">
+        <a class="button" href="https://join.slack.com/t/ceph-storage/shared_invite/zt-453jwr8e9-wu~QvC0jcH4oZBug3oMgdg">Subscribe</a>
+      </div>
+     <p class="p mb-0">
       Join thousands of operators and developers learning and discussing Ceph, the open-source de facto standard storage system.
       </p>
   </div>


anthonyeleven:
At one point a certain community member maintained a Slack <--> IRC bridge, but I haven't heard from him in quite a while and I think that no longer works.  I think we do want to encourage Slack over IRC, since we aren't living in 1986.

eameh-LF:
Agreed on Slack first. Reworked: the separate IRC card is gone; the Slack card stays primary and IRC is now a one-line pointer beneath it (#ceph and #ceph-devel on irc.oftc.net via webchat) noting that most discussion happens on Slack. No mention of the old bridge.

anthonymicmidd:
The IRC channels are not supported by the Ceph Foundation, but they are supported by the Ceph Community. Those servers can exist, but the Ceph Foundation should not promote them, mostly because the foundation doesn't have admin access to support community members.

anthonymicmidd:
> ## Summary
> The Connect page promoted Slack as the only chat option, but IRC on `irc.oftc.net` is still an official Ceph community channel; the docs [Get Involved page](https://docs.ceph.com/en/latest/start/get-involved/) lists `#ceph`, `#ceph-devel`, `#ceph-dashboard`, and `#cephfs`. This adds an IRC card below the Slack card with those channels, and the Team page's member IRC handles now have somewhere on the site that explains which network they refer to.
> 
> Also fixes a broken attribute on the Slack card paragraph: `<p lass="p mb-0">` → `<p class="p mb-0">` (the styling classes were silently not applied).

We should keep these separate. The docs pages can mention IRC, but it should not be mentioned on ceph.io

anthonymicmidd (commented):
Why is the Slack link being updated? Also, please remove the IRC redirect.

eameh-LF:
Understood: since the Foundation does not have admin access to the IRC servers, ceph.io should not promote them. All IRC content is now removed from this PR; what remains is purely the Slack card markup fix (the broken class attribute, heading level, and button wrapper). Retitled accordingly.

anthonyeleven:
Plus IRC is icky and kind of smells.

anthonyeleven (approved):

PR #1186, opened 2026-08-18
Add LongVan as Silver Member

bensternthal:
Add LongVan as Silver Member


neha-ojha (approved):
https://1186.ceph.io/en/foundation/members/

batrick:
Welcome!

PR #1192, opened 2026-08-23
Updating CFP Close Date for CD Zurich

anthonymicmidd:


anthonyeleven (approved):
One might argue for 11 September

PR #1193, opened 2026-08-26
Update Website To Reflect Trademark Transfer to LF

bensternthal:
## Move Ceph legal docs to the Linux Foundation

With the Ceph trademark transferred to the Linux Foundation, the site's legal pages
were still the Red Hat–era documents. This updates all three, plus the footer copyright.

### Trademark policy

The policy is now a PDF (March 2026, administered by LF Projects, LLC) rather than
prose we maintain. Rather than delete the page and chase inbound links, `/en/trademarks/`
stays put and its body becomes a short pointer:

- Adds `src/assets/pdfs/ceph-trademark-policy-march-2026.pdf`, served from
  `/assets/pdfs/…` via the existing `src/**/*.{csv,pdf,ods}` passthrough in `.eleventy.js`
  — no config change needed.
- Replaces the June 2016 Red Hat policy text with an intro, a prominent link to the PDF,
  a three-bullet orientation summary, and `trademarks@lfprojects.org` as the contact.
  The page states plainly that the PDF is authoritative.
- Front matter is untouched, so the URL, the footer position, the `/en/logos/` inline
  link, and the existing `/trademarks → /en/trademarks` redirect all keep working.

### Terms of use and privacy policy

Both now live off-site, so the pages are deleted rather than rewritten:

- `src/en/terms/index.md` (the ~41KB Red Hat ToS) and `src/en/privacy/index.md` removed.
- The footer support nav is generated from `collections.support`, so deleting the pages
  would have silently dropped both links. `site-footer.njk` now merges two external
  entries into that nav at the positions the pages held (`order: 2` and `3`), and the
  loop branches on an `external` flag to skip the `| url` filter and add
  `rel="noreferrer noopener" target="_blank"`.
- `redirects.j2` sends `^/(en/)?terms/?$` and `^/(en/)?privacy/?$` to the LF documents,
  covering both the legacy bare paths and the deleted pages' own URLs.

### Footer copyright

Now reads:

> Copyright © 2026 The Linux Foundation®. All rights reserved. The Linux Foundation has
> registered trademarks and uses trademarks.

This also fixes a pre-existing bug: the year came from `page.date.getFullYear()`, i.e. each
*page's* own date, so the 2014 blog post footer read "© 2014" and `/en/trademarks/` read
"© 2024". It now uses the existing `build.buildTime` global from `src/_data/build.js`,
so the year is consistent site-wide and rolls automatically on each build.

### Also updated

`src/en/logos/index.md` — two guideline bullets said "the Ceph project or Red Hat";
now "LF Projects, LLC", consistent with the new policy.

### Notes for reviewers

- **The PDF is the signed Docusign copy** — every page carries a `Docusign Envelope ID`
  header and it's titled "EXHIBIT C". Publishing it as-is was intentional, but flagging
  it in case a clean copy is preferred.
- **The redirects only take effect on deploy.** `redirects.j2` is an nginx template, not
  part of the static build — until the ansible deploy role runs, `/en/terms` and
  `/en/privacy` will 404 rather than redirect.
- The footer year now rolls on *build*, not wall-clock. If nothing triggers a rebuild
  after January 1st, the deployed site shows the previous year until the next deploy.

### Testing

`npm run build` from clean. Verified:

- `/en/trademarks/` renders the new copy; the PDF link resolves to
  `dist/assets/pdfs/ceph-trademark-policy-march-2026.pdf`.
- `dist/en/terms/` and `dist/en/privacy/` are no longer built; no `href="/en/terms"` or
  `href="/en/privacy"` remains anywhere in the output.
- The footer support nav renders in its original order — Code of conduct, Terms of use,
  Privacy policy, Trademarks, Logo Usage, Security — with the two LF links off-site, and
  `aria-current` still resolves on the remaining internal pages.
- All 1,398 pages carrying the footer render an identical copyright line.
- No `trademarks@redhat.com` / `tmlicense@redhat.com` references remain in `src/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code), with assistance from the human @bensternthal


anthonymicmidd (approved):
LGTM! Thanks for updating this, Ben

PR #1205, opened 2026-09-14
Update members page to use landscape

bensternthal:
Changes the members page to pull information from https://github.com/ceph/ceph-membership-landscape which pulls data directly from LFX. This means we will not have to make updates to this page manually, member changes will populate automatically. This pr includes the change and supporting changes such as a script to adjust the size of the iframe.


bensternthal:
Looks like the preview will not work, and will never work until the new csp policy is deployed. The same policy is used for production and preview. Once we do the merge once, preview here will work. Below is a screenshot of what this looks like locally
Screenshot 2026-09-14 at 11 56 26 AM

anthonymicmidd (approved):
LGTM! In Ben we trust!

PR #1208, opened 2026-09-14
Temporarily restore member logo grids until landscape CSP is deployed

bensternthal:
Turns out we need an ansible task to update the CSP policy for this to work. This commit preserves the CSP changes but reverts the iframe, we can then enable the iframe once the policy is update.


anthonymicmidd (approved):
LGTM

PR #1212, opened 2026-09-16
Restored Landscape Membership View, Now That CSP Is Adjusted

bensternthal:
Alrighty, ansible was run so the CSP policy should be updated, which also means this should work in preview!

This pr adds / embeds the landscape served from https://landscape.cephfoundation.org

This landscape updates every day with information from LFX meaning that membership changes, logo updates etc will all happen automatically.


anthonymicmidd (approved):
I worked! Looks great!

